PR #55, opened 2023-07-18
build: Whitelabeling.

Kelketek:
This PR adds some whitelabeling assistance to the repository.

The current configuration tracks the files which contain per-build customization in version control. This change takes files related to theming and configuration, removes them from direct version control, and then adds a makefile for setting up these initial files.

**Testing instructions**:

Follow the instructions in the README to set up a new copy of the repository and build it.

**Author notes and concerns**:

1. I'm not familiar with the XCode/iOS development workflow. There may be a more idiomatic way of accomplishing this.
2. I expect that most client developers will not actually want to customize all of these files. If the files are updated by upstream, they won't be copied over here-- unless the file does not yet exist in the target directory (I'm using rsync with --ignore-existing). It might be better to have some way of 'falling back' to default assets rather than freezing the version that exists upon initial setup. If this is needed, I'm not sure if the best way to do this would be to code some native swift that does this fallback work, or to create a script that pulls in assets from a theme directory in preparation for build. Advice here is requested.
3. I haven't yet looked at the Android version of this app-- presumably it would also need a similar modification. I'm not sure how tightly we're working to keep these in sync.

**Reviewers**
- [ ] @bradenmacdonald @Cup0fCoffee
- [ ] @raccoongang/educationx-app-ios-reviewers


volodymyr-chekyrta (changes requested):
Thank you for your contribution!

Could we make the Whitelabel step optional the same as the legacy edX app version?
https://github.com/openedx/edx-app-ios#whitelabel-script
As a required step, this can make it difficult for the community to run, understand and explore the project.

The [README.md](README.md) file contains two steps with the number 7.

Cup0fCoffee:
@Kelketek Could you help me understand the purpose of this change?

> The current configuration tracks the files which contain per-build customization in version control.

I assume this is a problem, because when building with custom theme/assets, you would have a dirty git state. But does it matter, if you store the custom theme and assets in a separate repo and copy them in before building? Are there any other reasons to why it's better to exclude the default theme and assets from the version control?

> This change takes files related to theming and configuration, removes them from direct version control, and then adds a makefile for setting up these initial files.

How does it improve the current workflow? I agree with @volodymyr-chekyrta that this adds an extra step for someone new to try this project, which makes it harder to try it out.

If you've changed anything, you can reset it back with something like `git reset --hard HEAD`, which seems to do the same as what you've tried to achieve with the whitelabeling, since it would reset all the changed files to the state stored in vc.

The only case I can think of when the current workflow is inconvenient, is if there are files like images or icons, that are part of the default theme, but are not in a custom theme you're trying to build with. In that case just copying the custom theme into the project would not be enough - you would also need to remove the files that are in the default theme, but not in yours. So having the theme and assets directories be empty by default makes it more convenient. Is this the usecase we want to optimize for? If so, is there another way we can do it?

Kelketek:
@Cup0fCoffee 

> I assume this is a problem, because when building with custom theme/assets, you would have a dirty git state. But does it matter, if you store the custom theme and assets in a separate repo and copy them in before building? Are there any other reasons to why it's better to exclude the default theme and assets from the version control?

The main reason is that it's a footgun. Say you maintain your institution's copy of the app, and you occasionally make changes you want to contribute upstream. You're likely to accidentally commit things you don't mean to when contributing upstream, or else have unnecessary conflicts from upstream on each pull.

We can expect that every institution that wants to build the app WILL need to customize several of the files I've factored out. Which ones will depend on the institution, and most will probably just make a fork of the repo and 'stuff everything in there', if that's the path the code suggests, setting them up for conflicts both when pulling changes and when contributing them. While they could all write individual scripts to pull the current version, copy over the assets, and then build, I think most will follow the path of least resistance and then this will cause headaches later.

It's much better if the client developer does not need to worry about their customizations (within certain bounds) having any conflict with upstream-- and the sooner we can institute such a path, the less effort will be needed later for anyone building with it today to migrate over.

> Could we make the Whitelabel step optional the same as the legacy edX app version?
https://github.com/openedx/edx-app-ios#whitelabel-script
As a required step, this can make it difficult for the community to run, understand and explore the project.

@volodymyr-chekyrta Let me see what I can do here. If I can find a way to do loading by fallback, then this should be easy enough, though I think I'll have to dive into swift a bit more.

Kelketek:
@volodymyr-chekyrta I've come to the conclusion that this is a much bigger project than I'm able to take in within my client's budget to get the DX right, so closing this.

PR #65, opened 2023-08-17
feat: Support web-based LMS OAuth.

Kelketek:
## Description

This pull request makes it possible to use the LMS's web login rather than the app's login form. This is necessary in the case that the client institution is using a custom authentication provider, and must piggy-back off the LMS's authentication flow to avoid writing a whole new custom login interface for these providers.

## Testing instructions

The best way to test this is to use the version of the app built for Yam education.

1. Register here: https://app.dev.yam-edu.com/
2. Then load up this PR and build it: https://github.com/yameducation/educationx-app-ios/pull/2
3. Load the app. You should immediately be brought to the login page.
4. Login, and verify the app's main interface loads.
5. Alternatively, hit Done to be brought back to a minimalist login page that will allow you to jump back into the authentication workflow.

## Deadline

None

## Author Concerns

One thing that's not working quite right in our client's application build is logout. This is because while the existing logout does, in fact, expire the LMS token, it doesn't log out the user from Auth0, the provider they're using. This means hitting the sign in button will log you right back in without prompting you for the password once more. It turns out that on the LMS, the logout button has a 'next' variable to a view that also logs the user out from Auth0. I've not yet determined if this is something we should also add as a tuneable variable-- having some custom logout URL. I'd have to see how that was implemented first.

My suspicion is that we should actually change what the LMS is doing rather than adding some additional logout action into the app code, but I'm not yet sure. cc @Cup0fCoffee

## Reviewers

* @bradenmacdonald
* RaccoonGang reviewer TBD

### DIFF
--- a/Authorization/Authorization/Presentation/Login/SignInView.swift
+++ b/Authorization/Authorization/Presentation/Login/SignInView.swift
@@ -19,163 +19,190 @@ public struct SignInView: View {
     
     @ObservedObject
     private var viewModel: SignInViewModel
+    private var navigationController: UINavigationController
     
-    public init(viewModel: SignInViewModel) {
+    public init(viewModel: SignInViewModel, navigationController: UINavigationController) {
         self.viewModel = viewModel
+        self.navigationController = navigationController
+    }
+    
+    public func webLogin() async {
+        await viewModel.login(viewController: self.navigationController)
     }
     
     public var body: some View {
-        ZStack(alignment: .top) {
-            VStack {
-                ThemeAssets.authBackground.swiftUIImage
-                    .resizable()
-                    .edgesIgnoringSafeArea(.top)
-            }.frame(maxWidth: .infinity, maxHeight: 200)
-            if viewModel.config.features.startupScreenEnabled {
-                VStack {
-                    Button(action: { viewModel.router.back() }, label: {
-                        CoreAssets.arrowLeft.swiftUIImage.renderingMode(.template)
-                            .backButtonStyle(color: .white)
-                    })
-                    .foregroundColor(Theme.Colors.styledButtonText)
-                    .padding(.leading, isHorizontal ? 48 : 0)
-                    .padding(.top, 11)
-                    
-                }.frame(maxWidth: .infinity, alignment: .topLeading)
-                    .padding(.top, isHorizontal ? 20 : 0)
+        if viewModel.forceWebLogin {
+            Task {
+                await webLogin()
             }
-            
-            VStack(alignment: .center) {
-                ThemeAssets.appLogo.swiftUIImage
-                    .resizable()
-                    .frame(maxWidth: 189, maxHeight: 54)
-                    .padding(.top, isHorizontal ? 20 : 40)
-                    .padding(.bottom, isHorizontal ? 10 : 40)
+        }
+        return ZStack(alignment: .top) {
+            if viewModel.forceWebLogin {
+                // Is there an idiomatic way of doing an early return here
+                // rather than using this big indented else clause?
+                // Using a return statement seems to break whatever magic
+                // happens here.
+            } else {
+                VStack {
+                    ThemeAssets.authBackground.swiftUIImage
+                        .resizable()
+                        .edgesIgnoringSafeArea(.top)
+                }.frame(maxWidth: .infinity, maxHeight: 200)
                 
-                ScrollView {
+                if viewModel.config.features.startupScreenEnabled {
                     VStack {
-                        VStack(alignment: .leading) {
-                            Text(AuthLocalization.SignIn.logInTitle)
-                                .font(Theme.Fonts.displaySmall)
-                                .foregroundColor(Theme.Colors.textPrimary)
-                                .padding(.bottom, 4)
-                            Text(AuthLocalization.SignIn.welcomeBack)
-                                .font(Theme.Fonts.titleSmall)
-                                .foregroundColor(Theme.Colors.textPrimary)
-                                .padding(.bottom, 20)
-                            
-                            Text(AuthLocalization.SignIn.emailOrUsername)
-                                .font(Theme.Fonts.labelLarge)
-                                .foregroundColor(Theme.Colors.textPrimary)
-                            TextField(AuthLocalization.SignIn.emailOrUsername, text: $email)
-                                .keyboardType(.emailAddress)
-                                .textContentType(.emailAddress)
-                                .autocapitalization(.none)
-                                .autocorrectionDisabled()
-                                .padding(.all, 14)
-                                .background(
-                                    Theme.Shapes.textInputShape
-                                        .fill(Theme.Colors.textInputBackground)
-                                )
-                                .overlay(
-                                    Theme.Shapes.textInputShape
-                                        .stroke(lineWidth: 1)
-                                        .fill(Theme.Colors.textInputStroke)
-                                )
-                            
-                            Text(AuthLocalization.SignIn.password)
-                                .font(Theme.Fonts.labelLarge)
-                                .foregroundColor(Theme.Colors.textPrimary)
-                                .padding(.top, 18)
-                            SecureField(AuthLocalization.SignIn.password, text: $password)
-                                .padding(.all, 14)
-                                .background(
-                                    Theme.Shapes.textInputShape
-                                        .fill(Theme.Colors.textInputBackground)
-                                )
-                                .overlay(
-                                    Theme.Shapes.textInputShape
-                                        .stroke(lineWidth: 1)
-                                        .fill(Theme.Colors.textInputStroke)
-                                )
-                            HStack {
-                                if !viewModel.config.features.startupScreenEnabled {
-                                    Button(AuthLocalization.SignIn.registerBtn) {
-                                        viewModel.trackSignUpClicked()
-                                        viewModel.router.showRegisterScreen()
-                                    }.foregroundColor(Theme.Colors.accentColor)
-                                    
-                                    Spacer()
-                                }
-                                    
-                                Button(AuthLocalization.SignIn.forgotPassBtn) {
-                                    viewModel.trackForgotPasswordClicked()
-                                    viewModel.router.showForgotPasswordScreen()
-                                }.foregroundColor(Theme.Colors.accentColor)
-                                    .padding(.top, 0)
-                            }
-                            
-                            if viewModel.isShowProgress {
-                                HStack(alignment: .center) {
-                                    ProgressBar(size: 40, lineWidth: 8)
-                                        .padding(20)
-                                }.frame(maxWidth: .infinity)
-                            } else {
-                                StyledButton(AuthLocalization.SignIn.logInBtn) {
-                                    Task {
-                                        await viewModel.login(username: email, password: password)
-                                    }
-                                }.frame(maxWidth: .infinity)
-                                    .padding(.top, 40)
-                            }
-                        }
-                        if viewModel.socialAuthEnabled {
-                            SocialAuthView(
-                                viewModel: .init(
-                                    config: viewModel.config
-                                ) { result in
-                                    Task { await viewModel.login(with: result) }
-                                }
-                            )
-                        }
-                        Spacer()
-                    }
-                    .padding(.horizontal, 24)
-                    .padding(.top, 50)
-                }.roundedBackground(Theme.Colors.background)
-                    .scrollAvoidKeyboard(dismissKeyboardByTap: true)
+                        Button(action: { viewModel.router.back() }, label: {
+                            CoreAssets.arrowLeft.swiftUIImage.renderingMode(.template)
+                                .backButtonStyle(color: .white)
+                        })
+                        .foregroundColor(Theme.Colors.styledButtonText)
+                        .padding(.leading, isHorizontal ? 48 : 0)
+                        .padding(.top, 11)
+                        
+                    }.frame(maxWidth: .infinity, alignment: .topLeading)
+                        .padding(.top, isHorizontal ? 20 : 0)
+                }
                 
-            }
-            
-            // MARK: - Alert
-            if viewModel.showAlert {
-                VStack {
-                    Text(viewModel.alertMessage ?? "")
-                        .shadowCardStyle(bgColor: Theme.Colors.accentColor,
-                                         textColor: Theme.Colors.white)
-                        .padding(.top, 80)
-                    Spacer()
+                VStack(alignment: .center) {
+                    ThemeAssets.appLogo.swiftUIImage
+                        .resizable()
+                        .frame(maxWidth: 189, maxHeight: 54)
+                        .padding(.top, isHorizontal ? 20 : 40)
+                        .padding(.bottom, isHorizontal ? 10 : 40)
+                    
+                    ScrollView {
+                        VStack {
+                            VStack(alignment: .leading) {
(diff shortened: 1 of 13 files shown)


Core/Core/Configuration/Config.swift
```diff
@@ -20,22 +21,23 @@ public class Config {
         URL(string: "\(baseURL.description)/privacy")
     }()
     
-    public let feedbackEmail = "support@example.com"
+    public let feedbackEmail = "contact@yam-edu.com"
```
bradenmacdonald, line 24: This line slipped in... I think you meant to exclude this.

Authorization/Authorization/Presentation/Login/SignInView.swift
```diff
@@ -16,137 +16,170 @@ public struct SignInView: View {
...
-            }.frame(maxWidth: .infinity, maxHeight: 200)
-            
-            VStack(alignment: .center) {
-                CoreAssets.appLogo.swiftUIImage
-                    .resizable()
-                    .frame(maxWidth: 189, maxHeight: 54)
-                    .padding(.vertical, 40)
+        if viewModel.forceWebLogin {
+            // On first load, we should bring the user right to the web login
+            // interface rather than showing this view.
```
bradenmacdonald, line 33: What do you think about this: instead of this boolean `forceWebLogin` which special-cases one login method, could we have a setting called `presumeLoginMethod` or `forceLoginMethod` which can optionally be set to one of the `LoginMethod` enum values?

Whether you call it `presumeLoginMethod` or `forceLoginMethod` depends on whether you allow the user to go "Back" and choose a different method, or if this is the only option available to them. Though I suppose the "force" could be achieved simply by disabling all but one of the login methods.

Kelketek: @bradenmacdonald

I'm not sure how I feel about this suggestion-- for two reasons. The first is that there's a rendering effect I'm concerned about here. The `forceWebLogin` method is there not only to bring the user to the web authentication page, but to make sure that, on initial load, the sign in page isn't drawn. There needs to be some way to signal to the view not to render the button or text if we're going to be opening the Safari browser to perform the login. Otherwise you get this visual glitch where you launch the app, see the sign-in page, and then the browser opens, which can be alarming and confusing-- I know that if I saw that as a user, I'd wonder if I pressed something I shouldn't or something happened I didn't want.

What you suggest could notionally be compatible with that first reason, but it would still require special casing the web login because, presumably, any other login method would have its interfaces written into the Swift view, and thus would need to be hidden if we're presuming the web login.

The second reason I'm not sure about your suggestion is that I'm not sure any other login methods, other than direct login, are actually viable outside of a web view, or wouldn't otherwise require some other complex changes to the UI. For example if you were to sign in with Google, you'd still need to visit the LMS login page to do the same dance we're doing for the client with Auth0. Same with any SAML provider, as far as I can tell-- the existing login view assumes that the user will be logging into the LMS using a post to Django directly, but nearly all external authentication methods will require visiting some intermediary web page and getting redirected back, requiring a Safari window, or else some tighter, complex iOS integration I'm unaware of.

For reference, I initially attempted to do the integration following Auth0's guide on integrating with iOS. However I realized as I was working with it that while I could authenticate with Auth0, the tokens it would provide me only worked with their servers directly. I couldn't get a token for the LMS without visiting the LMS and getting redirected from it to a login page provided by Auth0. There might be some other way to do it but it escapes me for now, and I suspect this is the way it's intended to be done. After all, all the MFEs at present just redirect to the LMS for login and then redirect you back, rather than presenting a streamlined login MFE.

Core/Core/Configuration/Config.swift
Kelketek: Thanks-- fixed!

Kelketek:
@bradenmacdonald Unmarking as WIP because although it probably can't be merged as-is, it does need upstream feedback from RaccoonGang before it can be moved forward.

cc @IvanStepanok @volodymyr-chekyrta If either of you has some time :) Thanks!

volodymyr-chekyrta (changes requested):
Hi, thank you for your efforts and input! 
And I apologize for the delayed review.
This feature looks quite complex to me, so let's take it step by step.

Could we please start with changing the target branch to develop, I know now it's a bit confusing compared to the legacy repos, but we would like to follow the git flow with the main and develop branches.
I'm going to describe this approach in the repos asap or we'll do it as part of the [OEP-64](https://github.com/openedx/open-edx-proposals/pull/496).

Authorization/Authorization/Presentation/Login/SignInViewModel.swift
```diff
@@ -29,20 +50,78 @@ public class SignInViewModel: ObservableObject {
...
+                    self.router.showMainScreen()
+                }
+                // Do your request
+            case .failure(let error):
+                self.webLoginAttempted = true
+                self.isShowProgress = false
+                self.errorMessage = error.localizedDescription
+            }
+        }
+    }
```
volodymyr-chekyrta, line 119: This function can be problematic for several reasons.
1. We are trying to encapsulate all navigation and ViewControllers logic inside of the Router entity.
This helps us to free ourselves from UIKit, navigation logic and be able to change the navigation framework at any time and write tests easily.
We could play with it and ask Router to provide us with configured SafariURLHandler instance or something like that.
2. I recommend converting the closure style function
`oauthswift!.authorize(withCallbackURL: rwURL, scope: "", state: "") { result in`
to async style function
`await oauthswift!.authorize(...)`  or `await oauthswift!.authorizeTask(...)`
we can reach it with some extension function and wrap `oauthswift!.authorize` to the [withCheckedThrowingContinuation](https://developer.apple.com/documentation/swift/withcheckedthrowingcontinuation(function:_:)) call.
3. It's better to remove `Task {}` creating from the async function as it makes it hard/impossible to test with Unit testing.
4. Last but not least, could you please extend the `SignInViewModelTests.swift` with tests for your new functions? This is a crucial part of the application's stability.
I highly recommend running the `generateAllMocks.sh` script before you start creating tests. This script creates all mocks for tests if its need.

If you need any help, hit me on Slack or email me 📪

Authorization/Authorization/Presentation/Login/SignInViewModel.swift
```diff
@@ -9,12 +9,33 @@ import Foundation
...
+        self.viewModel.markAttempted()
+    }
+}
 
 public class SignInViewModel: ObservableObject {
     
     @Published private(set) var isShowProgress = false
     @Published private(set) var showError: Bool = false
     @Published private(set) var showAlert: Bool = false
+    @Published private(set) var webLoginAttempted: Bool = false
```
volodymyr-chekyrta, line 34: View doesn't use the variable `webLoginAttempted`.
Maybe we could remove `@Published` and make it private?

Authorization/Authorization/Presentation/Login/SignInViewModel.swift
```diff
@@ -29,20 +50,78 @@ public class SignInViewModel: ObservableObject {
...
                 router: AuthorizationRouter,
                 analytics: AuthorizationAnalytics,
+                config: Config,
                 validator: Validator) {
         self.interactor = interactor
         self.router = router
         self.analytics = analytics
+        self.config = config
         self.validator = validator
+        self.webLoginAttempted = false
```
volodymyr-chekyrta, line 72: Is it needed to initialize `webLoginAttempted` with `false` on init if the variable is already initialized `var webLoginAttempted: Bool = false`?

volodymyr-chekyrta (changes requested):

Authorization/Authorization/Presentation/Login/SignInView.swift
```diff
@@ -16,137 +16,170 @@ public struct SignInView: View {
...
+            //
+            // If that login fails or the user escapes back, they'll be brought
+            // back to the view where any error message will be shown.
+            Task {
+                await webLogin()
+            }
+        }
```
volodymyr-chekyrta, line 40: In my mind, a more SwiftUI way will put this code to the View `.onAppear {}`
What do you think about it?

volodymyr-chekyrta:
> Today I attempted to work on the tests, but it looks like I can't run them properly on non-Apple silicon. The tests rely on autogenerated mocks which need to be refreshed in order to update them and run, and the Podfile installs a dependency for generating these-- SwiftyMocky, which is prebuilt for Apple Silicon only.
> 
> I attempted to compile it for `x86_64` and it did compile, but when it ran on the needed directory, it failed compiling on some dependency that it only pulls in and builds at runtime. Searching around for the cause didn't yield much useful info-- I'd have to spelunk way down the rabbit hole to fix it, so it might just be best to find someone else with Apple Silicon who can complete the test portion or ask RaccoonGang for help here.

@Kelketek if you have a non-Apple Silicon chip, try this
https://github.com/MakeAWishFoundation/SwiftyMocky#2-installing-swiftymocky-cli
Just run SwiftyMocky commands with Mint (tested on Intel i5)

Authorization/Authorization/Presentation/Login/SignInView.swift
Kelketek: @volodymyr-chekyrta I'm quite new to Swift, SwiftUI, and its idioms. This is my first Swift code, in fact :)

So, I couldn't tell you for sure. But I've looked up the docs, and I think you're probably right! I'll add this.

Authorization/Authorization/Presentation/Login/SignInViewModel.swift
Kelketek: Nope-- that got left in when I was trying a few things. I'll remove it.

Authorization/Authorization/Presentation/Login/SignInViewModel.swift
Kelketek: Sure!

Authorization/Authorization/Presentation/Login/SignInViewModel.swift
Kelketek: Thanks, @volodymyr-chekyrta ! This is great feedback. I'll ping you as I make progress on it, as this appears to be the bulk of the changes to be made.

e0d:
@Kelketek  was a product ticket created for this work?  We're pushing to have all PRs tie back to a product ticket that is, ideally, prioritized before the development happens.

openedx-webhooks:
Thanks for the pull request, @Kelketek! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

Kelketek:
@e0d No, no product ticket was created for this work. It was discovered as necessary for a client when preparing a copy of the app for them. We did get the build working for them with these customizations, but due to other snags there is no anticipation of continuing development on this at this time.

Still, this is likely to be a fairly important thing to have for any team which is using anything other than direct auth to the platform (so, pretty much all SSO users). [It could be rolled into this existing ticket, which this fix would solve.](https://github.com/orgs/openedx/projects/56?pane=issue&itemId=39291363)

To whoever that person is, there's one other bug that showed up later: It doesn't look like this implementation actually lets you log out. Likely, what is needed is the ability to clear all Safari state for the app in order to perform the logout. However I'm unfamiliar with the ecosystem to know how to do this off-hand (this is my first foray into Swift.)

@xitij2000 worked on a similar fix for Android. I believe it has the same features and limitations.

Kelketek:
@e0d Looks like we've got another client who needs this, so I'm continuing work on it. I've created an issue here: https://github.com/openedx/openedx-app-ios/issues/212

Let me know if you need anything else on that front.

marcotuts (approved):
Product review only 👍

Kelketek:
Hi @volodymyr-chekyrta ! We're currently discussing internally how to handle the budget for finishing this PR. For context, we've now had two different clients who have needed this but haven't yet secured enough from either of them to get it across the finish line. We're trying to figure out what to do about it and how to get buy-in for the finishing work.

I'll update you as soon as I have a prediction on when we'll be able to rebase this and address your notes.

volodymyr-chekyrta:
> Hi @volodymyr-chekyrta ! We're currently discussing internally how to handle the budget for finishing this PR. For context, we've now had two different clients who have needed this but haven't yet secured enough from either of them to get it across the finish line. We're trying to figure out what to do about it and how to get buy-in for the finishing work.
> 
> I'll update you as soon as I have a prediction on when we'll be able to rebase this and address your notes.

Makes sense to me 👍 
Please let us know if we can help you from the engineering or any other perspective.

Kelketek:
Had a conversation with Zemeel Hassan from Blend-ed who is considering taking over this PR on behalf of one of their clients. [Recording here](https://drive.google.com/file/d/1LgUIHppogI-hnR-GLcb7KApfFe-LTsyj/view?usp=drivesdk), where we walk a little bit through the code and some of the issues that he's running into with the rebase (and which I'd previously run into on previous rebases)

zameel7:
> Had a conversation with Zemeel Hassan from Blend-ed who is considering taking over this PR on behalf of one of their clients. [Recording here](https://drive.google.com/file/d/1LgUIHppogI-hnR-GLcb7KApfFe-LTsyj/view?usp=drivesdk), where we walk a little bit through the code and some of the issues that he's running into with the rebase (and which I'd previously run into on previous rebases)

Hey! I have started working on rebasing it to the develop branch. I have some doubts which I'll have to clear prior to drafting the PR. How should I proceed?
@Kelketek @volodymyr-chekyrta

Kelketek:
> How should I proceed?

@zameel7 Go ahead and ask them here! I'm sure @volodymyr-chekyrta will be able to answer your questions about creating the replacement PR and can likely answer some questions about the changes that have been made since this original version.

volodymyr-chekyrta:
Hi @zameel7 @Kelketek, please let me know if you have any questions. I'll do my best to help you.

The only doubt I have is that we need to double-check that we do not duplicate the behavior introduced at https://github.com/openedx/openedx-app-ios/pull/447.
Perhaps we should extend the logic presented in https://github.com/openedx/openedx-app-ios/pull/447 rather than adding another SAML approach.

Kelketek:
@zameel7 The merged PR that @volodymyr-chekyrta just linked might actually solve what this PR was attempting to do incidentally. Could you build that version of the app and check if it works for you? If it does, we could close this PR and be done here :)

zameel7:
> @zameel7 The merged PR that @volodymyr-chekyrta just linked might actually solve what this PR was attempting to do incidentally. Could you build that version of the app and check if it works for you? If it does, we could close this PR and be done here :)

Sure, let me check that. I wonder how I missed the context of that PR😀

Kelketek:
@zameel7 That's what I don't know. I would suspect it might, since I would think the SAML auth would turn into a more native grant on behalf of the platform once activated via web auth, but the only way to know would be to try it out. It might not.

But even if not, starting from that branch will already give you a leg up in setting up external auth.

zameel7:
I tested it and it works. For the SSO URL, I gave the URL that the OAuth swift used to build for authorise and for the SSO FINISHED URL, I gave the redirect callback url to the mobile app(I have a doubt here).

e0d:
@Kelketek There are some requested changes and branch conflicts currently.  Is there a plan to address?

Kelketek:
Hi @e0d,

No, the original client declined to move forward, and from my understanding, this functionality has been implemented in another PR.

e0d:
@volodymyr-chekyrta is it your understanding this was implemented in another PR?  I think the Android version of this work was merged, but not the IOS version, which isn't a great place to be.

PR #104, opened 2023-10-11
Added support of JWT tokens.

volodymyr-chekyrta:
By default, access tokens are now in JWT format.
The type of access token can be changed in the [Core.Config](https://github.com/openedx/openedx-app-ios/blob/feature/add_support_for_jwt_tokens/Core/Core/Configuration/Config.swift) class.

Supported token types are represented by the [TokenType](https://github.com/openedx/openedx-app-ios/blob/feature/add_support_for_jwt_tokens/Core/Core/Configuration/Config.swift#L36) enum.
```
enum TokenType: String {
    case jwt = "JWT"
    case bearer = "BEARER"
}
```

### DIFF
--- a/Core/Core/Configuration/Config.swift
+++ b/Core/Core/Configuration/Config.swift
@@ -11,6 +11,7 @@ public class Config {
     
     public let baseURL: URL
     public let oAuthClientId: String
+    public let tokenType: TokenType = .jwt
     
     public lazy var termsOfUse: URL? = {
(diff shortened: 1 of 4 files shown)


mumer92 (approved):

openedx-webhooks:
Although this pull request is already merged,  I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks again for your contribution.

PR #140, opened 2023-10-31
App configuration protocol

eyatsenkoperpetio:


Core/Core/Configuration/Configurable/ConfigMock.swift
```diff
@@ -78,7 +78,19 @@ public class ConfigMock: Configurable {
...
+
+    public var isGoogleSignInEnabled: Bool {
+        false
+    }
+
+    public var isFacebookSignInEnabled: Bool {
+        false
+    }
+
+    public var isMicrosoftSignInEnabled: Bool {
```
saeedbashir, line 93: @eyatsenkoperpetio How are you planning to use Facebook and other authentication provider keys in the configuration? Are you planning to directly add them to info.plist?

eyatsenkoperpetio: we should to discuss about it, can we configure authentication providers without plist like Google?

Core/Core/Configuration/Configurable/Configurable.swift
```diff
@@ -0,0 +1,55 @@
...
+//
+
+import Foundation
+import FirebaseCore
+import Swinject
+
+public enum Apps {
+    case openEdx
+    case edX
+}
```
saeedbashir, line 15: @eyatsenkoperpetio What are we trying to achieve here by creating different enums?

Core/Core/Configuration/Configurable/ConfigMock.swift
eyatsenkoperpetio: we should extend this configuration file

saeedbashir: I'm afraid we can't do that with all the authentication providers, for example, Facebook, Microsoft needs some keys to be added to the info.plist

We are achieving this by reading the values from the config file at compile time through a script and  adding them to the info.plist

Our config was look like this 
```
FACEBOOK:
    ENABLED: true
    FACEBOOK_APP_ID: 'facefoob_app_id'
    CLIENT_TOKEN: 'facefoob_client_token'
```

saeedbashir:
I didn't get a chance to look into the config PR throughly, from the overview, I got an idea that we are trying to achieve two things here:

1. Navigational management: Any provider can not only change the theme(yet to be implemented) but can also change the whole UI, for example if any provider wants to provide a different sign-in screen, one can override the base router and use their own sign-in screen.

2. Config files Management: We can have multiple config files, at the moment one for the openedX and other for the edX.

Here are my thoughts on the above points:
1.  Navigational management: The approach is good but I'm not sure whether we need this at this point (start of the project). And maintaining multiple routers can be confusing.
2.  I'm not fully sure whether there will be only two enums edX and OpenedX config OR all the providers will be maintaining or adding there own enum values, but if thats the case this will not be a good approach because there will be a lot of enums. 

If we are planning to have only two enums, edX and OpenedX then adding a new target might be a more relevant and friendly solution here. Because after adding a new target we will be able to configure the edX app more quickly and easily because the edX target will be having all the resources and stuff separately like, fonts, colors, logo, images and even identifier and team setting etc.

cc @miankhalid @touchapp

saeedbashir:
And instead of baking all the data in the configuration file, how about having it in a separate file like in JSON or in YAML. YAML would be nice because we can add comments in YAML.

```
API_HOST_URL: 'https://courses.edx.org'

API_URL_VERSION: 
    BLOCKS: 'v2'
    REGISTRATION: 'v2'
    ENROLLMENTS: 'v2'
        
PLATFORM_NAME: 'edX'
PHONETIC_PLATFORM_NAME: 'E D X'

AGREEMENT_URLS:
    PRIVACY_POLICY_URL: 'https://www.edx.org/webview/edx-privacy-policy'
    COOKIE_POLICY_URL: 'https://www.edx.org/webview/edx-privacy-policy/cookies'
    DATA_SELL_CONSENT_URL: 'https://www.edx.org/webview/edx-privacy-policy/do-not-sell-my-personal-data'
    TOS_URL: 'https://www.edx.org/webview/edx-terms-service'
    EULA_URL: 'https://www.edx.org/webview/eula'
    SUPPORTED_LANGUAGES: ['es'] #en is defalut language
```

PR #141, opened 2023-10-31
fix: add support for login via username

mumer92:
This PR adds support for username and email to be used to login to app.

- User can login using username
- User can login using email

Simulator Screenshot - iPhone 15 Pro 17 - 2023-11-01 at 18 00 46


saeedbashir (changes requested):

Authorization/Authorization/Presentation/Login/SignInViewModel.swift
```diff
@@ -50,12 +50,12 @@ public class SignInViewModel: ObservableObject {
     
     @MainActor
     func login(username: String, password: String) async {
-        guard validator.isValidEmail(username) else {
-            errorMessage = AuthLocalization.Error.invalidEmailAddress
+        guard !username.isEmpty else {
```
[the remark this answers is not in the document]
saeedbashir: @mumer92 Can you look into this as well.

saeedbashir:
@mumer92 It's not giving an error for an empty `username`.

saeedbashir (approved):
LGTM 👍

mumer92:
@volodymyr-chekyrta waiting for your review on this.

Core/Core/View/Validator.swift
```diff
@@ -9,24 +9,23 @@ import Foundation
 
 public class Validator {
     
-    private let emailRegEx = "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,64}"
+    private let emailRegEx = ".+@.+\\..+"
```
volodymyr-chekyrta, line 12: What is the reason for changing this regex expression?

mumer92: reverted it back to original.

Authorization/Authorization/en.lproj/Localizable.strings
```diff
@@ -9,14 +9,15 @@
 "SIGN_IN.LOG_IN_TITLE" = "Sign in";
 "SIGN_IN.WELCOME_BACK" = "Welcome back! Please authorize to continue.";
 "SIGN_IN.EMAIL" = "Email";
```
volodymyr-chekyrta, line 11: Looks like this string is no longer in use and should be removed.

PR #148, opened 2023-11-03
Feature/in app review system

IvanStepanok:
https://github.com/openedx/openedx-app-ios/assets/128456094/040b0a3b-19ec-468f-996a-ffa462437f8b

https://github.com/openedx/openedx-app-ios/assets/128456094/c1e3c897-6875-41f3-8d11-19701f1b307d


openedx-webhooks:
Thanks for the pull request, @IvanStepanok! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

sdaitzman:
This looks great to me! It would be ideal to collect feedback directly in-app, but that requires some backend communication and is probably not in-scope for now.

Two questions about the interface:

- What currently triggers this dialog? Is it timing-based? Could it be triggered only when a video is not actively playing, maybe immediately after a video?
- Is the mail client picker that appears at the bottom custom? I believe iOS supports changing default mail apps now, so I think it would be better to use the user's chosen mail client without prompting them to choose one

mumer92:
@rnr one feedback i have to to make a separate directory and place 3rd party libraries we are using in project to that directory. in that way, we will be able to see what external dependencies we are using, it will have better organization of of code base. what are your thoughts?

rnr:
> @rnr one feedback i have to to make a separate directory and place 3rd party libraries we are using in project to that directory. in that way, we will be able to see what external dependencies we are using, it will have better organization of of code base. what are your thoughts?

@mumer92 do you mean something else except Pods and Swift Packages (I believe these two Sources are organised well already)? Thank you

mumer92:
> > @rnr one feedback i have to to make a separate directory and place 3rd party libraries we are using in project to that directory. in that way, we will be able to see what external dependencies we are using, it will have better organization of of code base. what are your thoughts?
> 
> @mumer92 do you mean something else except Pods and Swift Packages (I believe these two Sources are organised well already)? Thank you

My understanding is `ThirdPartyMailClient` external library? If it is, we should create a separate folder on root level and place files to that folder, e.g. `ExternalLibraries/ThirdPartyMailClient` and same with other third party files that are added to project.

volodymyr-chekyrta:
> This looks great to me! It would be ideal to collect feedback directly in-app, but that requires some backend communication and is probably not in-scope for now.
> 
> Two questions about the interface:
> 
> * What currently triggers this dialog? Is it timing-based? Could it be triggered only when a video is not actively playing, maybe immediately after a video?
> * Is the mail client picker that appears at the bottom custom? I believe iOS supports changing default mail apps now, so I think it would be better to use the user's chosen mail client without prompting them to choose one

Yes, it's based on timing here, it appears at 99.9% of the video.

After chatting with @marcotuts, we thought it's a good idea to give users a choice. 
Most people have the Apple client by default, but there's a bunch who prefer Gmail.

We can make changes if needed.

volodymyr-chekyrta:
> > > @rnr one feedback i have to to make a separate directory and place 3rd party libraries we are using in project to that directory. in that way, we will be able to see what external dependencies we are using, it will have better organization of of code base. what are your thoughts?
> > 
> > 
> > @mumer92 do you mean something else except Pods and Swift Packages (I believe these two Sources are organised well already)? Thank you
> 
> My understanding is `ThirdPartyMailClient` external library? If it is, we should create a separate folder on root level and place files to that folder, e.g. `ExternalLibraries/ThirdPartyMailClient` and same with other third party files that are added to project.

That's not the library we're going to update in the future.
Here it's just a part of the App Review feature, so it's located there.

Core/Core/View/Base/AppReview/AppReviewViewModel.swift
```diff
@@ -0,0 +1,143 @@
...
+            withAnimation(Animation.bouncy(duration: 0.2)) {
+                showSelectMailClientView = true
+            }
+        } else {
+            openMailClient(ThirdPartyMailClient.systemDefault)
+        }
+    }
+    
+    func requestReview() {
+        SKStoreReviewController.requestReview()
```
rnr, line 98: Hello @IvanStepanok 
As I know SKStoreReviewController can be showed three times per year only
https://developer.apple.com/documentation/storekit/requesting_app_store_reviews
All other times 'Rate Us' button won't work, right?
Thank you

marcotuts:
A new notes - can we trigger the rating once the video ends instead of at 99.99% - this way the interrupt isn't near the end of an activity but right as it ends. 

Similarly on the note about rating visibility, I can review the code next but ensuring this doesnt appear more than a certain number of times is important. the requirements only specified that the modal should appear at most ever 3 months, but if you chose to rate the app we should actually have a timer that is less frequent / perhaps yearly? Thoughts on this change in logic / addition to the logic @IvanStepanok ? 

I wasn't familiar with the 3 limit max @rnr  thanks for f lagging that.

volodymyr-chekyrta:
> @volodymyr-chekyrta `ThirdPartMailer` is available as a library via pod, i believe this is the link to github project. `https://github.com/vtourraine/ThirdPartyMailer`
> 
@mumer92,
We know the library is available via pods, yet it is not currently under active development. 
Its most recent update was on April 18, 2022, and it does not utilize the latest version of Swift. 
Therefore, we have opted to implement only the necessary features instead of relying on this framework.

volodymyr-chekyrta:
> A new notes - can we trigger the rating once the video ends instead of at 99.99% - this way the interrupt isn't near the end of an activity but right as it ends.

99.99% of the time, this serves as a workaround for the YouTube player, which does not consistently report full 100% progress. 
However, we've conducted tests on a wide array of videos and found that typically, the final second marked by 99.99% progress does not contain any significant content.

volodymyr-chekyrta:
@marcotuts, we didn't see major releases coming more than three times a year, but that's a good point you've raised with @rnr. 
We'll add an additional check to the logic. 
What do you think works better - limiting it to once every four months or every six months?

sdaitzman:
> Yes, it's based on timing here, it appears at 99.9% of the video.
> 
> After chatting with @marcotuts, we thought it's a good idea to give users a choice. Most people have the Apple client by default, but there's a bunch who prefer Gmail.
> 
> We can make changes if needed.

That makes sense, I think it's reasonable to include the choice. In that case, this looks good to me on the design side!

saeedbashir:
> We know the library is available via pods, yet it is not currently under active development.

@volodymyr-chekyrta I guess what Umer is saying is that there should be clarity in what code we wrote and what code we are using from a third-party or external source.

And for that clarity, he was suggesting to create a folder named (ExternalLibraries/3rdPartyLibraries) at root in the framework where we are using a third-party library and place the 3rd party code in that folder instead of having it in the subdirectory. For example, in the rating feature, a folder named ExternalLibraries would be created in the Core framework and `ThirdPartyMailClient.swift` placed in that folder.

volodymyr-chekyrta:
> > We know the library is available via pods, yet it is not currently under active development.
> 
> @volodymyr-chekyrta I guess what Umer is saying is that there should be clarity in what code we wrote and what code we are using from a third-party or external source.
> 
> And for that clarity, he was suggesting to create a folder named (ExternalLibraries/3rdPartyLibraries) at root in the framework where we are using a third-party library and place the 3rd party code in that folder instead of having it in the subdirectory. For example, in the rating feature, a folder named ExternalLibraries would be created in the Core framework and `ThirdPartyMailClient.swift` placed in that folder.

I understand, but I'm not quite sure why we need to create a new folder and relocate these files into it for this specific feature.
I strongly believe that this is redundant in this particular situation.

saeedbashir:
> but I'm not quite sure why we need to create a new folder and relocate these files into it for this specific feature

There are multiple reasons why we are suggesting moving 3rd party libraries to a central location:

- Most of the open-source libraries are available under the MIT license; in given case, [ThirdPartyMailer](https://github.com/vtourraine/ThirdPartyMailer) is available under MIT license which means we should attribute the author 
- We don't need to review the third-party code, but if there isn't any indication of 3rd-party code then we will end up reviewing that code as well
- It will be much easier to list down all the third-party libraries that we are using.

volodymyr-chekyrta:
@marcotuts, upgraded logic in the last commit, so now the review popup only shows once every four months.

Core/Core/View/Base/AppReview/AppReviewViewModel.swift
```diff
@@ -0,0 +1,160 @@
...
```
mumer92, line 114: app has minimum target of iOS 15, this api was deprecated in iOS 14, how about using the updated api?

Core/Core/View/Base/AppReview/AppReviewViewModel.swift
```diff
@@ -0,0 +1,160 @@
...
+        storage.lastReviewDate = Date()
+    }
+    
+    func openMailClient(_ with: ThirdPartyMailClient) {
+        
+        let osVersion = UIDevice.current.systemVersion
+        let appVersion = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? ""
```
mumer92, line 121: `Bundle.main.infoDictionary?["CFBundleShortVersionString"]` is being used at multiple places, i think we should centralize this somewhere, what are your thoughts?

Core/Core/View/Base/AppReview/AppReviewView.swift
```diff
@@ -0,0 +1,139 @@
...
+                .onTapGesture {
+                    presentationMode.wrappedValue.dismiss()
+                }
+            if viewModel.showSelectMailClientView {
+                SelectMailClientView(clients: viewModel.clients, onMailTapped: { client in
+                    viewModel.openMailClient(client)
+                })
+            } else {
+                VStack(spacing: 20) {
+                    if viewModel.state == .thanksForFeedback || viewModel.state == .thanksForVote {
```
mumer92, line 34: how about using `switch` here instead of if else?

mumer92 (changes requested):

Core/Core/View/Base/AppReview/AppReviewView.swift
IvanStepanok: Good idea, done!

Core/Core/View/Base/AppReview/AppReviewViewModel.swift
IvanStepanok: Replace to `@Environment(\.requestReview) var requestReview` and move logic from AppReviewViewModel to AppReviewView.

Core/Core/View/Base/AppReview/AppReviewViewModel.swift
IvanStepanok: We'll need to think about this as part of the next refactoring.

mumer92 (approved):

volodymyr-chekyrta (approved):

openedx-webhooks:
@IvanStepanok 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-app-ios&entry.867055334=https://github.com/openedx/openedx-app-ios/pull/148&entry.1484655318=https://github.com/IvanStepanok&entry.752974735=2023-11-03+11:55&entry.1917517419=2023-11-17+11:51&entry.2133058324=Yes) so we can improve your experience in the future.

PR #137, opened 2023-10-26
feat: course dates

mumer92:
This PR adds support for Course Dates. https://github.com/openedx/openedx-app-ios/issues/85

This is preliminary PR with the Course Dates feature. Designs are TBD.

Screenshots:
Screenshot 2023-10-31 at 11 56 18 PM
Screenshot 2023-10-31 at 11 56 35 PM

_Note_
Not included in this ticket:
- Calendar integration
- PLS (shift due dates)
- Linking of items on the dates page to specific assessments within the app

PR #149, opened 2023-11-06
chore: course dates feature

mumer92:
This is continuity of https://github.com/openedx/openedx-app-ios/pull/137
to address feedback, as original PR was merged.

This PR also fixes edge case for the Today block and siblings for Today block, small color fixes to Circle and bit of refactoring.

### DIFF
--- a/Course/Course/Presentation/Dates/CourseDatesView.swift
+++ b/Course/Course/Presentation/Dates/CourseDatesView.swift
@@ -35,7 +35,7 @@ public struct CourseDatesView: View {
                     }
                 } else if let courseDates = viewModel.courseDates, !courseDates.courseDateBlocks.isEmpty {
                     CourseDateListView(viewModel: viewModel, courseDates: courseDates)
-                        .padding(.top)
+                        .padding(.top, 10)
                 }
             }
             if viewModel.showError {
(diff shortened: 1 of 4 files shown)


Course/Course/Presentation/Dates/CourseDatesView.swift
```diff
@@ -35,6 +35,7 @@ public struct CourseDatesView: View {
                     }
                 } else if let courseDates = viewModel.courseDates, !courseDates.courseDateBlocks.isEmpty {
                     CourseDateListView(viewModel: viewModel, courseDates: courseDates)
+                        .padding(.top)
```
volodymyr-chekyrta, line 38: Please set a specific value for `.padding(.top)`. 
You might use `.padding(.top, 20)` as an example.

mumer92: @volodymyr-chekyrta i have added padding value to this, but i think this should be handled by the parent or maybe we should handle this by adding NavigationTtile to the child view. This same padding/overlapping issue will occur on the handouts tab.

Course/CourseTests/Presentation/Unit/CourseDateViewModelTests.swift
```diff
@@ -5,12 +5,12 @@
 //  Created by Muhammad Umer on 10/24/23.
 //
 
-import SwiftyMocky
+import SwiftUI
```
saeedbashir, line 8: Tests are running fine without this import `SwiftUI` .

saeedbashir (commented):
Just a small comment otherwise working fine

saeedbashir (approved):
LGTM 👍

volodymyr-chekyrta (approved):
LGTM

PR #158, opened 2023-11-13
feat: configuration management

mumer92:
**Project Configuration Management**

This PR introduces a way to integrate and manage config files for iOS project.

- Build Phase Script Integration: Adds a script to Build Phase of Xcode. It calls Xcode build phase run script, which takes care of virtual environment and installing dependencies and calls a python script `process_config.py` with `$CONFIGURATION` and `scheme_mappings` argument.
- Python Script for Configuration: Utilizes process_config.py for:
  - Adding essential keys to `Info.plist` (e.g., Facebook, Microsoft keys).
  - Creating`GoogleServices.plist` with Firebase keys.
  - Generating `config.plist` from `ios.yaml` and shared.yaml.

Inside `Config.swift` parsing and populating relevant keys and classes are done, e.g. `AgreementConfig.swift` and `FirebaseConfig.swift` 

**Get started:** 
Edit a `config_settings.yaml` in the `default_config` folder.
It should contain data as follows

```
config_directory: '{path_to_confiig_folder}'
config_mapping:
  prod: 'prod'
  stage: 'stage'
  dev: 'dev'
# These mappings are configurable, e.g. dev: 'prod_test'
```

- `config_directory` provides path of config directory
- `config_mappings` provides provides mappings that can be utilized to map Xcode build scheme map to a defined folder within config directory and it will be referenced.

Config files are separated out in 2 files

`ios.yaml` and `shared.yaml`

both configs should be placed under a folder which will be referenced inside `config_settings.yaml` inside of the key `config_mappings:{config_directory_name}` 

inside `config_directory`, `mappings.yaml` which contains yaml files that will be used to read available yaml files and processed inside script.

it is structure will be as follows:

```
ios:
    files:
      - {file_one.yaml}
      - {file_two.yaml}
```

- `ios.yaml` will contain config data specific to iOS, e.g. `Firebase` keys, `Facebook` keys, etc.
- `shared.yaml` will contain config data that is shared, e.g. `API_HOST_URL`, `OAUTH_CLIENT_ID`, `TOKEN_TYPE`, etc

**Future Support:**
- To add config related to some other service, create a class, e.g. `ServiceNameConfig.swift` to be able to populate related fields
- Create `extension` to Config.swift to be able to add the newly created service as a varable to main Config.
- If needed, make a protocol to be reference inside the scope of  `ConfigProtocol` so that config is available using `ConfigProtocol` service.

e.g.
```
private let key = "KEY"
extension Config {
    public var serviceNameConfig: ServiceNameConfig {
        return ServiceNameConfig(dictionary: self[key] as? [String: AnyObject] ?? [:])
    }
}
```

**Note**
This PR also updates `FirebaseCrashlytics` build phase
Updated build phase script extracts `googleAppID` from the newly generated `GoogleService-Info.plist` and runs the crashlytics script with the provided id. 

**Examples of config files:**

`ios.yaml`
```
OAUTH_CLIENT_ID: ''

FIREBASE:
  ENABLED: true
  API_KEY: "testApiKey"
  BUNDLE_ID: "testBundleID"
  CLIENT_ID: "testClientID"
  DATABASE_URL: "https://test.database.url"
  GCM_SENDER_ID: "testGCMSenderID"
  GOOGLE_APP_ID: "testGoogleAppID"
  PROJECT_ID: "testProjectID"
  REVERSED_CLIENT_ID: "testReversedClientID"
  STORAGE_BUCKET: "testStorageBucket"
  ANALYTICS_SOURCE: "firebase"
  CLOUD_MESSAGING_ENABLED: true

MICROSOFT:
  ENABLED: true
  APP_ID: "microsoftAppID"
```

`shared.yaml`
```
API_HOST_URL: "https://www.example.com"
FEEDBACK_EMAIL_ADDRESS: "example@mail.com"
TOKEN_TYPE: "JWT"

AGREEMENT_URLS:
  PRIVACY_POLICY_URL: "https://www.example.com/privacy"
  TOS_URL: "https://www.example.com/tos"

# Features
WHATS_NEW_ENABLED: false
```

`default_config` directory is added to project to get an idea on how to write config yaml files.


OpenEdX.xcodeproj/project.pbxproj
```diff
@@ -375,6 +373,24 @@
 			shellScript = "diff \"${PODS_PODFILE_DIR_PATH}/Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [ $? != 0 ] ; then\n    # print error to STDERR\n    echo \"error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\" >&2\n    exit 1\nfi\n# This output is used by Xcode 'outputs' to avoid re-running this script phase.\necho \"SUCCESS\" > \"${SCRIPT_OUTPUT_FILE_0}\"\n";
 			showEnvVarsInLog = 0;
 		};
+		DB97C0542B002EF00035C36F /* Process Config */ = {
```
rnr, line 376: @mumer92 Now to switch between Stg and Prod version I need to edit config.yml file manually, right? What do think about adding parameter to process_config.py and call it with own config.yml file (every file will point out to particular folder prod/stage/prod_test for example in existing 'edx-mobile-config' repo) for every Scheme? This way you just select another Scheme in Xcode to switch between environments.

OpenEdX/Environment.swift
rnr, line 1: @mumer92  We need to change 'FirebaseCrashlytics' build phase if we got rig Environment.swift file. Thank you

Core/Core/Configuration/Combine/Config/FirebaseConfig.swift
```diff
@@ -0,0 +1,105 @@
...
+    
+    public var isAnalyticsSourceSegment: Bool {
+        return analyticsSource == AnalyticsSource.segment
+    }
+    
+    public var isAnalyticsSourceFirebase: Bool {
+        return analyticsSource == AnalyticsSource.firebase
+    }
+    
+    public var firebaseOptions: FirebaseOptions? {
```
rnr, line 80: @mumer92  Do we need this firebaseOptions variable if FirebaseApp.configure() uses generated .plist now? Thank you

OpenEdX.xcodeproj/project.pbxproj
rnr: Screenshot 2023-11-13 at 17 16 04
This is how we can modify 'Process Config' Build Phase to check current Scheme. 
If user want to modify config he could point any Scheme to own config files path.
It looks similar how to determine 'path' in config.yaml you suggested but it could be done only once and don't need to change 'environment' in config.yaml every time you would jump from prod to stg.
And this way saves Schemes which are critical for openEdx how @volodymyr-chekyrta said.

rnr: Instead "Unknown configuration" we can call script with some 'default' path

mumer92: i have removed `process_config.sh` in favour of directly running script from build phase, passing `$CONFIGURATION` as argument to script and also updated `process_config.py` to be able to handle `$CONFIGURATION`. if configuration is not provided, it runs with the config provided inside `config.yaml`

OpenEdX/Environment.swift
mumer92: i have update `FirebaseCrashlytics` to be able to extract `googleAppID` from the `Google-Service.plist` and use that in the calling the script

Core/Core/Configuration/Combine/Config/FirebaseConfig.swift
mumer92: i have added back the Firebase initialization that was using `FirebaseOptions`

mumer92:
@rnr @volodymyr-chekyrta @saeedbashir I have updated the implementation to use Xcode schemes that are currently i use. i have updated the description with these changes.

OpenEdX.xcodeproj/project.pbxproj
rnr: @mumer92 in this case if user wants to add new Scheme he needs to edit script file, right? Also script strongly determines folders name as 'prod', 'prod-test' and 'stage' (what if user wants to have 'sandbox'?). If we could create new Scheme and define which folder is attached to it in one place (in Xcode build phase) it's better from my view. What are your thoughts?
Also one question is what is pros to have this 'config.yaml' file - why 'directory' can't be a part of 'path' script's  parameter in build phase too?
Thank you

mumer92: We generally had 3 configs in edX project, but you are welcome to add them to project. The idea is to have to do minimal to no change, in that way, you will just chose scheme and config is applied for you, based on scheme.

in `config.yaml` the mappings are provided as follows
```
config_mapping: 
  dev: 'dev'
  prod: 'prod'
  stage: 'stage'
```

the idea for providing the mappings it to have a flexibility, e.g.
```
config_mapping: 
  dev: 'prod_test'
  prod: 'prod'
  stage: 'stage'
```

if someone has `prod_test` instead of `dev`, they would be able to specify the structure here and mappings are read from the script and script looks for the files based on the value, 
e.g. if you have 3 folders, `prod - prod_test - stage`, you will provide these names as mappings here and script then maps the directory name to the appropriate key, in this case, if you run Xcode with `DebugDev` scheme, script will read file mappings and treat them as directory name and find yaml files inside that directory and loads them in script.

for reference, please see following:
https://github.com/openedx/openedx-app-ios/blob/7b09d047af8827a1f6cd7a9e9e440fb2464c94dc/process_config.py#L214C1-L222C20

rnr: Hello @mumer92 
Flexability is exactly what I'm saying about. When I want to add one more Scheme 'Sandbox' I need to to config.yaml and add mapping there (one more question what exists still is what if this 'sandbox' config doesn't live in defined there 'config_directory'?). After that I need to edit python script by adding configuration name and mapping value.
Instead of all this stuff I'm showing how we can do it without config.yaml and python script editing
Screenshot 2023-11-15 at 11 16 33
If we need to add new Scheme (or point a Scheme out to other config) you just add one line here:
`"DebugSandbox" | "RelesaseSandbox" ) python process_config.py "$PATH_TO_CONFIG/sandbox" ;;` (and here PATH_TO_CONFIG can be any - default or user's)
So my question is why python script should construct path to config from config.yaml and from mapping if it can get path as parameter and work with it?
@volodymyr-chekyrta please share your thoughts here

mumer92: @rnr I dont think much of developer effort would be required in this case,
e.g. there are 3 configs provided as default on project, and if developer wants to add another config, they can either modify this function to handle configuration and provide mappings, or break down the case:

```
def get_configuration_name(configuration, config_mapping):
    if configuration in ["DebugDev", "ReleaseDev"]:
        return config_mapping.get("dev")
    elif configuration in ["DebugStage", "ReleaseStage"]:
        return config_mapping.get("stage")
    elif configuration in ["DebugProd", "ReleaseProd"]:
        return config_mapping.get("prod")
    else:
        return None
```

or e.g. they can update the `config.yaml` to update the mapping, and i dont think either apporach would take much of developer effort, could be done in a minute or 2.

rnr: @mumer92 
Why did we go from discussing flexibility and usability to developer effort? Let's please get back to discussing the advantages and disadvantages of approaches.
I don't like the fact that changing/adding a schema in Xcode causes the need to go and fix a python script - in my opinion it's not about flexibility. Why we create a rigid connection between scheme names and python script?
My second question is why do we need to create an additional config.yaml file if we already have a build phase dedicated to configuration, which can contain the schema-config mapping? What is pros of this file?
Thank you

saeedbashir: I think multiple cases are being discussed here and are getting confused with each other.

@rnr How about if we can finalize this PR first and then discuss it in more detail over a call to make it more mature? I'll be looking into it with @mumer92. Scenario, we have a large community that has already published their apps with their configurations (with a specific / existing format). We want to have something where there will be minimal change while shifting to the new codebase. 

We don't have the luxury of freedom here to follow anything. So given the circumstances, we have to come up with the best possible solution, which might not be the best solution.

cc @miankhalid @touchapp

touchapp:
@volodymyr-chekyrta would like your opinion here. What @rnr is proposing is more elegant and straight forward that involves less steps. I suggest we go with that. The concerns about community raised by @saeedbashir can be addressed if we make clear documentation on how to configure. @rnr can add these changes but want to make sure we are aligned.

saeedbashir:
> @volodymyr-chekyrta would like your opinion here. What @rnr is proposing is more elegant and straight forward that involves less steps. I suggest we go with that. The concerns about community raised by @saeedbashir can be addressed if we make clear documentation on how to configure. @rnr can add these changes but want to make sure we are aligned.

I and @rnr had a discussion on it, and we agreed on a structure. Today we'll be making the changes and then this PR will be ready for final pass.

mumer92:
@rnr @volodymyr-chekyrta i have made the changes that we agreed on, the PR is ready for review.

Core/Core/Configuration/Combine/Config/Config.swift
```diff
@@ -0,0 +1,140 @@
...
+    }
+    
+    func bool(for key: String) -> Bool {
+        return value(for: key) as? Bool ?? false
+    }
+}
+
+extension Config: ConfigProtocol {
+    public var baseURL: URL {
+        return URL(string: string(for: ConfigKeys.baseURL.rawValue)!)!
```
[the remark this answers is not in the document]
rnr: I understand baseURL is important part and we need to have this
maybe it can be something like
```    
    public var baseURL: URL {
        if let string = string(for: ConfigKeys.baseURL2.rawValue),
            let url = URL(string: string) {
            return url
        }
        fatalError("Can't find baseURL in Config!")
    }
```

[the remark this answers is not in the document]
mumer92: i used convention that was being followed in project, i think we can use guard let here and return `URL(string "")` instead of doing `fataError`, what are your thoughts?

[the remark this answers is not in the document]
mumer92: i am not in favour of `fataError()`, i think app should handle issue gracefully.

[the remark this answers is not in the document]
rnr: @mumer92 code we have now (with force unwrapping) will just crash if baseURL is not present in config.
URL(string: "") also returns nil and will crash with force unwrapping
I understand app can't work without baseURL - we need to try to show what is the problem exactly (fatalError show a message at least). It could be some Log etc. Just provide issue description.
I'm open to any options without force unwrapping.
What do you think?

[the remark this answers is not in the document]
mumer92: what are your thoughts on using:
```
public var baseURL: URL {
        guard let urlString = string(for: ConfigKeys.baseURL.rawValue),
              let url = URL(string: urlString) else {
            return URL(string: "")!
        }
        return url
    }
```

i believe we can do the same with `oAuthClientId`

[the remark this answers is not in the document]
rnr: URL(string: "")! produces crash
why this Crash is better than ?Crash with message"? )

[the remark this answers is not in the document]
mumer92: alright i will add `fatalError("Can't find baseURL in Config!")`, in the mean time you can continue with other changes for the review.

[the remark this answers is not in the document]
rnr: all other looks good for me. good work @mumer92 !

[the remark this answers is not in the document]
mumer92: @rnr i have updated it. please have a look.

rnr (dismissed):

volodymyr-chekyrta:
Let's move config files from the Combine folder, probably it's been placed here by mistake.
image

Core/Core/Configuration/Combine/Config/ConfigTests.swift
volodymyr-chekyrta, line 1: The test files must be located inside the CoreTests target

volodymyr-chekyrta:
We should place the Combine framework files back in it's folder 
image

OpenEdX/AppDelegate.swift
```diff
@@ -31,8 +31,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
     ) -> Bool {
         
-        if BuildConfiguration.shared.firebaseOptions.apiKey != "" {
-            FirebaseApp.configure(options: BuildConfiguration.shared.firebaseOptions)
+        if Config.shared.firebase.enabled,
```
volodymyr-chekyrta, line 34: This is a dependency violation. Config should be used from the DI service.

Core/Core/Configuration/Combine/Config/Config.swift
```diff
@@ -0,0 +1,147 @@
...
+    case tokenType = "TOKEN_TYPE"
+    case feedbackEmailAddress = "FEEDBACK_EMAIL_ADDRESS"
+    case environmentDisplayName = "ENVIRONMENT_DISPLAY_NAME"
+    case platformName = "PLATFORM_NAME"
+    case organizationCode = "ORGANIZATION_CODE"
+    case appstoreID = "APP_STORE_ID"
+}
+
+public class Config {
+    public static let shared = Config()
```
volodymyr-chekyrta, line 38: Dependency violation. Should use singleton only from the DI service.

Core/Core/Configuration/Combine/Config/Config.swift
```diff
@@ -0,0 +1,147 @@
...
+    
+    private convenience init() {
+        self.init(properties: [:])
+        loadAndParseConfig()
+    }
+    
+    private func loadAndParseConfig() {
+        guard let path = Bundle.main.path(forResource: "config", ofType: "plist"),
+              let data = try? Data(contentsOf: URL(fileURLWithPath: path)),
+              let dict = try? PropertyListSerialization.propertyList(from: data, options: [], format: nil) as? [String: Any] else { return }
```
volodymyr-chekyrta, line 54: Please fix the Swiftlint warning.
`Line Length Violation: Line should be 120 characters or less; currently it has 140 characters (line_length)`

-> 
```
guard let path = Bundle.main.path(forResource: "config", ofType: "plist"),
              let data = try? Data(contentsOf: URL(fileURLWithPath: path)),
              let dict = try? PropertyListSerialization.propertyList(
                from: data,
                options: [],
                format: nil) as? [String: Any] else { return }
```
or
```
guard let path = Bundle.main.path(forResource: "config", ofType: "plist"),
              let data = try? Data(contentsOf: URL(fileURLWithPath: path)),
              let dict = try? PropertyListSerialization.propertyList(
                from: data,
                options: [],
                format: nil
              ) as? [String: Any] else {
            return
        }
```

volodymyr-chekyrta:
I played around with configurations a bit and it works amazing 🚀🚀🚀

Core/Core/Configuration/Combine/Config/ConfigTests.swift
mumer92: i think this file is already part of `CoreTests` target,

Screenshot 2023-11-21 at 1 47 39 PM

mumer92:
@volodymyr-chekyrta i have addressed the feedback.

volodymyr-chekyrta (approved):

volodymyr-chekyrta:
Green light from me 🟢 🚀
I see we are still waiting for approval from @saeedbashir  
@saeedbashir, Could you please provide a review/approval?

Core/Core/Configuration/Combine/Config/ConfigTests.swift
volodymyr-chekyrta: I got it. It confused me because it's located in the Core folder instead of the CoreTests folder.

Can we move it to the CoreTests folder to keep consistency with other modules, like AuthorizationTest, DiscoveryTests, etc.?

PR #182, opened 2023-11-29
feat: Open edX app theming capability improvements (Option 2)

saeedbashir:
In this PR, I've made the changes that will support app theming. Below are the changes that I made:

Multiple colors and images
I've broken down the xcassets folder of Core Framework and move the configureables (most likely those will change for theming of openedX app) and moved them into a new framework `Theme`. From now on, Core framework will contain all the images, icons, colors, etc, that won't change with theming of the app, like button icons, nav bar icons, and bottom bar icons. Likewise, `Theme Assets` will contain all the changeable assets.

Fonts

I've added an extra layer between the fonts and the theme class. The font names were hardcoded in the theme class, which made it a bit hard to change the fonts in a new theme. I've added a json file for key: name pair and a parser around it. Now the font change will only require the replacement of fonts JSON and font_file, and there won't be any code changes.

Note: It's the second PR for making the theming easier. Another PR with different approach is already created and available here: https://github.com/openedx/openedx-app-ios/pull/173


saeedbashir:
@volodymyr-chekyrta @rnr @eyatsenkoperpetio I've created this PR with a different approach. Hopefully, this approach will work because we are already following the framework approach, and I've created another framework, `Theme` and moved the changeable things into the framework. Can you guys please take a look into it and let me know about this approach?

rnr:
> @volodymyr-chekyrta @rnr @eyatsenkoperpetio I've created this PR with a different approach. Hopefully, this approach will work because we are already following the framework approach, and I've created another framework, `Theme` and moved the changeable things into the framework. Can you guys please take a look into it and let me know about this approach?

I'm ok with both approach where we break down Assets on static and configurable since we need to change it with whitelabel script (but even without dividing whitelabel script can change assets just needs more configs perhaps). Using 'framework' way looks more structural for me. I'm waiting for @volodymyr-chekyrta inputs here.
Thank you.

miankhalid:
@touchapp would love to have your thoughts on this approach as well.

volodymyr-chekyrta:
@saeedbashir makes sense to me 🙌
Thank you!

One concern.
I see that the `Theme` framework is not linked as a framework to any other module. 
I am trying to understand why it's a building project now, but something might break it in the future.
Screenshot 2023-11-30 at 11 25 00

We probably need to link frameworks to each other in the same way as we do with other modules.
Screenshot 2023-11-30 at 11 28 47

saeedbashir:
> We probably need to link frameworks to each other in the same way as we do with other modules.

What benefit will we get after linking the theme framework to any other framework, like to Core framework?

volodymyr-chekyrta:
> > We probably need to link frameworks to each other in the same way as we do with other modules.
> 
> What benefit will we get after linking the theme framework to any other framework, like to Core framework?

There are no benefits, that just shouldn't work that way. 
Frameworks are essentially independent mini-projects. 
For instance, you can't utilize a pod dependency unless you've incorporated it into the project.

So, the question arises - how does this work, especially since the prior version of XCode would have thrown an error in this situation 🤔

saeedbashir:
@volodymyr-chekyrta Can you please look into this PR, there is a big change-set because of structual change. If we don't merge it early, most probably every merge will result in a conflict.

volodymyr-chekyrta:
> @volodymyr-chekyrta Can you please look into this PR, there is a big change-set because of structual change. If we don't merge it early, most probably every merge will result in a conflict.

I'm currently waiting for reviews from @rnr and @eyatsenkoperpetio, but I'll take a closer look today to speed up this process 🤝

rnr (approved):

saeedbashir:
> @saeedbashir, the project is building to the device, but the Archive build is not working. Could you please check it?

It is fixed; the debug statement was missing for Preview.

volodymyr-chekyrta:
> > @saeedbashir, the project is building to the device, but the Archive build is not working. Could you please check it?
> 
> It is fixed; the debug statement was missing for Preview.

Thank you! I'll check a little later

OpenEdX/AppDelegate.swift
```diff
@@ -12,6 +12,7 @@ import FirebaseCore
 import FirebaseAnalytics
 import FirebaseCrashlytics
 import Profile
+import Theme
```
volodymyr-chekyrta, line 15: Seems like there might be a redundant import here.

volodymyr-chekyrta (approved):

PR #185, opened 2023-11-30
feat: navigate to course component from course dates tab

mumer92:
This PR adds support for navigating to First Component ID, which is available from a Block inside Course Dates, to Course Unit screen.

https://github.com/openedx/openedx-app-ios/assets/960241/7c190627-4eeb-410b-9b49-7c7bf04fe3ea

PR #189, opened 2023-12-05
feat: navigate to course component from course dates tab

saeedbashir:
This PR adds support for navigating to the First Component ID, which is available from a block inside Course Dates, to the Course Unit screen.

https://github.com/openedx/openedx-app-ios/assets/5606473/baf1cd75-26c8-4dba-9b6b-ebed8168ebe9

**Note:** I didn't have write permissions on Umer's fork, so I'm taking over the PR and creating a new PR from the existing PR (https://github.com/openedx/openedx-app-ios/pull/185).
In the new PR, I've created it on the latest develop branch and fixed the case where on every hyperlink click blocks API was being hit.

### DIFF
--- a/OpenEdX/Router.swift
+++ b/OpenEdX/Router.swift
@@ -307,6 +307,33 @@ public class Router: AuthorizationRouter,
         navigationController.pushViewController(controller, animated: true)
     }
     
+    public func showCourseComponent(
+        componentID: String,
+        courseStructure: CourseStructure) {
+            courseStructure.childs.enumerated().forEach { chapterIndex, chapter in
+                chapter.childs.enumerated().forEach { sequentialIndex, sequential in
+                    sequential.childs.enumerated().forEach { verticalIndex, vertical in
+                        vertical.childs.forEach { block in
+                            if block.id == componentID {
+                                DispatchQueue.main.async { [weak self] in
+                                    guard let self = self else { return }
+                                    self.showCourseUnit(courseName: courseStructure.displayName,
+                                                        blockId: block.blockId,
+                                                        courseID: courseStructure.id,
+                                                        sectionName: sequential.displayName,
+                                                        verticalIndex: verticalIndex,
+                                                        chapters: courseStructure.childs,
+                                                        chapterIndex: chapterIndex,
+                                                        sequentialIndex: sequentialIndex)
+                                    return
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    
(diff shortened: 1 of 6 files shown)


saeedbashir:
@rnr You have already approved the old PR that Umer created. It would be nice if you could take a look at it and approve it as well. Thanks

saeedbashir:
@shafqat-muneer Here is the PR for dates hyperlink, could you please take a look?

rnr (dismissed):

OpenEdX/Router.swift
```diff
@@ -307,6 +307,33 @@ public class Router: AuthorizationRouter,
...
+                                    guard let self = self else { return }
+                                    self.showCourseUnit(courseName: courseStructure.displayName,
+                                                        blockId: block.blockId,
+                                                        courseID: courseStructure.id,
+                                                        sectionName: sequential.displayName,
+                                                        verticalIndex: verticalIndex,
+                                                        chapters: courseStructure.childs,
+                                                        chapterIndex: chapterIndex,
+                                                        sequentialIndex: sequentialIndex)
+                                    return
```
shafqat-muneer, line 328: I believe we can eliminate the use of `return` in this context.

shafqat-muneer (changes requested):

OpenEdX/Router.swift
volodymyr-chekyrta: We may need to use `return` to terminate all four cycles, but regardless, it should be placed outside the `DispatchQueue.main.async` closure.

OpenEdX/Router.swift
```diff
@@ -307,6 +307,33 @@ public class Router: AuthorizationRouter,
...
+                                DispatchQueue.main.async { [weak self] in
+                                    guard let self = self else { return }
+                                    self.showCourseUnit(courseName: courseStructure.displayName,
+                                                        blockId: block.blockId,
+                                                        courseID: courseStructure.id,
+                                                        sectionName: sequential.displayName,
+                                                        verticalIndex: verticalIndex,
+                                                        chapters: courseStructure.childs,
+                                                        chapterIndex: chapterIndex,
+                                                        sequentialIndex: sequentialIndex)
```
volodymyr-chekyrta, line 327: Each argument should be written on its own line 🙏
https://google.github.io/swift/#function-calls

shafqat-muneer (approved):
LGTM 👍

volodymyr-chekyrta (approved):

Issue #186, opened 2023-12-02
[iOS] "Open in Browser" button is partially hidden in Landscape mode


sergeymomot:
[Link to content](https://learning-raccoonapis.sandbox.edx.org/course/course-v1:edx+btest01+2023_BT/block-v1:edx+btest01+2023_BT+type@sequential+block@376b3893fe2a45f4b3fec7335b3c8640/block-v1:edx+btest01+2023_BT+type@vertical+block@43570fda10404cc1b7f948c35647adcb) 

"Open in Browser" button for unsupported content is partially hidden in landscape mode on devices with small screen (iPhone Xr for example)

<img width="440" alt="8761df4a-056c-47c3-83ee-b79c10f93af1" src="https://github.com/openedx/openedx-app-ios/assets/2964930/8761df4a-056c-47c3-83ee-b79c10f93af1">

PR #216, opened 2023-12-20
Open in browser button is partially hidden in landscape mode

rnr:
Small fix for https://github.com/openedx/openedx-app-ios/issues/186 bug
In landscape mode, the screen size is not enough to display the "Open in Browser" button if the content has a title. I've placed this "Unknown" view inside a scroll view.

https://github.com/openedx/openedx-app-ios/assets/37253/fde106b8-8a81-4b07-b2d5-b685909feb8c


volodymyr-chekyrta (approved):

saeedbashir (approved):
LGTM 👍

PR #232, opened 2024-01-12
fix: navigation bar colors

rnr:
This PR fixes bug with Navigation Bar background and title colors in dark mode. 
295306543-aa89d0b1-3501-43a3-a317-d6f82b2fe70f

By some reason calculated UIColor doesn't work for navigation bar (like `Theme.Colors.textPrimary.uiColor()`) and I added UIColors struct into Theme.

After fix Navigation Bar looks like this:
37253/6793dc56-1831-44c6-af30-11fc26d28cd2

### DIFF
--- a/Theme/Theme/Theme.swift
+++ b/Theme/Theme/Theme.swift
@@ -91,19 +91,17 @@ public struct Theme {
         }
     }
     
+    // Use this structure where the computed Color.uiColor() extension is not appropriate.
     public struct UIColors {
         public private(set) static var textPrimary = ThemeAssets.textPrimary.color
         public private(set) static var accentColor = ThemeAssets.accentColor.color
-        public private(set) static var background = ThemeAssets.background.color
 
         public static func update(
             textPrimary: UIColor = ThemeAssets.textPrimary.color,
-            accentColor: UIColor = ThemeAssets.accentColor.color,
-            background: UIColor = ThemeAssets.background.color
+            accentColor: UIColor = ThemeAssets.accentColor.color
         ) {
             self.textPrimary = textPrimary
             self.accentColor = accentColor
-            self.background = background
         }
     }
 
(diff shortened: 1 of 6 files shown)


forgotvas (approved):

volodymyr-chekyrta (dismissed):

Theme/Theme/Theme.swift
```diff
@@ -91,6 +91,22 @@ public struct Theme {
...
+            textPrimary: UIColor = ThemeAssets.textPrimary.color,
+            accentColor: UIColor = ThemeAssets.accentColor.color,
+            background: UIColor = ThemeAssets.background.color
+        ) {
+            self.textPrimary = textPrimary
+            self.accentColor = accentColor
+            self.background = background
+        }
+    }
+
```
saeedbashir, line 109: @rnr Why it's needed to create a new struct UIColors when we already have `uiColor()` extension written on `Colors`?

Even if we want to define categorical UIColors, I guess it would be nice to have all the values that Colors have and also it would be nice if you can remove `uiColor()` extension so we will have a single source. Thoughts?

saeedbashir:
@rnr I'm only able to replicate the issue when I revert the changes of `CourseContainerView` i.e., when I remove the background color.

Could you please check it again to see if you are able to reproduce the issue by using `uiColor()` in `UINavigationController` `viewWillLayoutSubviews`.

rnr:
> @rnr I'm only able to replicate the issue when I revert the changes of `CourseContainerView` i.e., when I remove the background color.
> 
> Could you please check it again to see if you are able to reproduce the issue by using `uiColor()` in `UINavigationController` `viewWillLayoutSubviews`.

@saeedbashir 
yes I'm able
For `navigationBar.titleTextAttributes = [.foregroundColor: Theme.Colors.textPrimary.uiColor()]`:
Screenshot 2024-01-16 at 08 57 47
To see problem with `navigationBar.backIndicatorImage` you need to change `accentColor` to non-blue color (because it gets default color when error):
Screenshot 2024-01-16 at 09 03 07

forgotvas (approved):

saeedbashir (approved):

Issue #224, opened 2024-01-09
[iOS] Hide one screen from "What's new"


sergeymomot:
Add possibility for hide "Learning site switching" screen from "What's new" info

<img width="280" alt="493e47f8-4b96-4404-84e4-c57fe6039503" src="https://github.com/openedx/openedx-app-ios/assets/2964930/493e47f8-4b96-4404-84e4-c57fe6039503">

PR #245, opened 2024-01-22
Whats new page hidding

rnr:
PR for https://github.com/openedx/openedx-app-ios/issues/224
Added FF `WHATS_NEW_IMAGE_OR_TITLE_TO_SKIP_PAGE` - it can contain the image name or page title from the What's new JSON file. Pages with such image name or title will be skipped in the What's new function.
To test you can use this config `https://github.com/edx/edx-mobile-config/tree/anton/whatsnew` or add this to `shared.yaml` manually like:
```yaml
UI_COMPONENTS:
    WHATS_NEW_IMAGE_OR_TITLE_TO_SKIP_PAGE: 'Learning Site Switching'
```
The `process_config.py` script has also been changed, it will now try to merge dictionaries from different files (e.g. `iOS.yaml` and `shared.yaml`) if both files have the same section with children in them

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,4 +116,3 @@ venv/
 Podfile.lock
 config_settings.yaml
 default_config/
-default_config/config_settings.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -116,3 +116,4 @@ venv/
 Podfile.lock
 config_settings.yaml
 default_config/
+default_config/config_settings.yaml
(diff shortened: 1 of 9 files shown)


forgotvas (dismissed):

saeedbashir:
@rnr As discussed over the huddle, this PR isn't needed as every version will be having it's own copy of whatsnew messages and images.

rnr:
Looks like this should be resolved differently. Thanks to @saeedbashir for the clarification.

PR #263, opened 2024-01-30
Abstract layer for Push Notifications and Deep Linking

rnr:
Draft PR to review and discuss an abstract implementation of Push notifications and deep link processing 
### What was done
- added `PushNotificationsManager` and `DeepLinkManager` to DI
- added protocols for `PushNotificationsProvider` and `PushNotificationsListener`
- added protocol for `DeepLinkService`
- added `DeepLink` and `PushLink` classes, which are used to redirect processes from deep link and from push notifications (DeepLink does not currently contain all possible types)
- added `configurations` for `Braze` and `Branch` services
- added empty implementation for `FCM` and `Braze` providers and listeners
- added empty implementation for `Branch` deep linking service

### How `PushNotificationsManager` works
`PushNotificationsManager` creates arrays of providers and listeners given the configuration (which service(s) are enabled). When any AppDelegate method is called (e.g. registerWithDeviceToken, getRemoteNotification), PushNotificationsManager translates that call to providers or listeners. Each provider/listener has its own initialization and processing functions. A listener can also call the `DeepLinkManager` to process the received notification data and trigger a redirect to the appropriate view.

### How `DeepLinkManager` works
`DeepLinkManager` contains one service that will handle deep links. The type of redirect will be determined by the `DeepLinkType` from the `DeepLink` or `PushLink` classes. The service processes the URL and, if it is a supported deep link, calls `DeepLinkManager.processDeepLink()` with parameters and the manager performs the redirect.


saeedbashir:
@rnr The suggested approach is very close to what we have on the production app, and the approach is modular and can adopt new providers very easily. We can go with this approach.

forgotvas (dismissed):

OpenEdX/AppDelegate.swift
```diff
@@ -125,4 +137,32 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         window?.rootViewController = RouteController()
     }
     
+    // Push Notifications
+    func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
+        if let pushManager = Container.shared.resolve(PushNotificationsManager.self) {
```
saeedbashir, line 142: How about using guard let here?

OpenEdX/AppDelegate.swift
```diff
@@ -125,4 +137,32 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            pushManager.didRegisterForRemoteNotificationsWithDeviceToken(deviceToken: deviceToken)
+        }
+    }
+    func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
+        if let pushManager = Container.shared.resolve(PushNotificationsManager.self) {
```
saeedbashir, line 147: How about using guard let here and same for the following push notifications methods?

OpenEdX/Managers/DeepLinkManager/DeepLinkManager.swift
```diff
@@ -0,0 +1,71 @@
...
+    func configureWith(launchOptions: [UIApplication.LaunchOptionsKey: Any]?)
+    func handledURLWith(app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any]) -> Bool
+}
+
+class DeepLinkManager {
+    private var service: DeepLinkService?
+    
+    // Init manager
+    public init(config: ConfigProtocol) {
+        self.service = self.serviceFor(config: config)
```
saeedbashir, line 22: The DeeplinkManager can register just one service Branch service. How about configuring it to register multiple services thoughts?

OpenEdX/Managers/DeepLinkManager/DeepLinkManager.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+    
+    // check if service is added (means enabled)
+    var serviceEnabled: Bool {
+        service != nil
+    }
+    
+    // Configure services
+    func configureDeepLinkService(launchOptions: [UIApplication.LaunchOptionsKey: Any]?) {
+        if let service = service {
```
saeedbashir, line 40: How about using guard let here?

OpenEdX/Managers/DeepLinkManager/DeepLinkManager.swift
```diff
@@ -0,0 +1,71 @@
...
+        open url: URL,
+        options: [UIApplication.OpenURLOptionsKey: Any] = [:]
+    ) -> Bool {
+        if let service = service {
+            return service.handledURLWith(app: app, open: url, options: options)
+        }
+        return false
+    }
+    
+    // This method process push notification with the link object
```
saeedbashir, line 57: We might need to add a more descriptive comment here because it's a bit odd to have push notification process in the deeplink manager. Or maybe we can create another class that will handle data that received from push and deep link and router the learner to the desired screen.

OpenEdX/Managers/PushNotificationsManager/PushNotificationsManager.swift
```diff
@@ -0,0 +1,96 @@
...
+import UIKit
+import Swinject
+
+public protocol PushNotificationsProvider {
+    func registerWithDeviceToken(deviceToken: Data)
+    func didFailToRegisterForRemoteNotificationsWithError(error: Error)
+}
+
+protocol PushNotificationsListener {
+    func shouldListenNotification(userinfo: [AnyHashable: Any]) -> Bool
```
saeedbashir, line 19: Is this method necessary? as if a push listener is register it should be listening to the push notifications. Or there will be other cases?

OpenEdX/Managers/PushNotificationsManager/Listeners/FCMListener.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  FCMListener.swift
+//  OpenEdX
+//
+//  Created by Anton Yarmolenka on 24/01/2024.
+//
+
+import Foundation
+
+class FCMListener: PushNotificationsListener {
```
saeedbashir, line 10: I guess it will be implementing the `didReceiveRemoteNotification` as well.

OpenEdX/Managers/PushNotificationsManager/Providers/BrazeProvider.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Created by Anton Yarmolenka on 24/01/2024.
+//
+
+import Foundation
+
+class BrazeProvider: PushNotificationsProvider {
+    func registerWithDeviceToken(deviceToken: Data) {
+        
+    }
+    func didFailToRegisterForRemoteNotificationsWithError(error: Error) {
```
saeedbashir, line 14: A newline between the two methods will be nice.

OpenEdX/Managers/PushNotificationsManager/Providers/FCMProvider.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Created by Anton Yarmolenka on 24/01/2024.
+//
+
+import Foundation
+
+class FCMProvider: PushNotificationsProvider {
+    func registerWithDeviceToken(deviceToken: Data) {
+        
+    }
+    func didFailToRegisterForRemoteNotificationsWithError(error: Error) {
```
saeedbashir, line 14: A newline between the two methods will be nice.

OpenEdX/Managers/PushNotificationsManager/PushNotificationsManager.swift
```diff
@@ -0,0 +1,96 @@
...
+//
+
+import Foundation
+import Core
+import UIKit
+import Swinject
+
+public protocol PushNotificationsProvider {
+    func registerWithDeviceToken(deviceToken: Data)
+    func didFailToRegisterForRemoteNotificationsWithError(error: Error)
```
saeedbashir, line 15: It would be nice if both methods followed the same naming convention either with did or without did.

OpenEdX/Managers/PushNotificationsManager/PushNotificationsManager.swift
```diff
@@ -0,0 +1,96 @@
...
+protocol PushNotificationsListener {
+    func shouldListenNotification(userinfo: [AnyHashable: Any]) -> Bool
+    func didReceiveRemoteNotification(userInfo: [AnyHashable: Any])
+}
+
+extension PushNotificationsListener {
+    func didReceiveRemoteNotification(userInfo: [AnyHashable: Any]) {
+       guard let dictionary = userInfo as? [String: Any], shouldListenNotification(userinfo: userInfo) else { return }
+       let link = PushLink(dictionary: dictionary)
+       if let deepLinkManager = Container.shared.resolve(DeepLinkManager.self) {
```
saeedbashir, line 27: How about moving it with guard let `dictionary`?

OpenEdX/Managers/PushNotificationsManager/PushNotificationsManager.swift
```diff
@@ -0,0 +1,96 @@
...
+}
+
+class PushNotificationsManager {
+    private var providers: [PushNotificationsProvider] = []
+    private var listeners: [PushNotificationsListener] = []
+    
+    // Init manager
+    public init(config: ConfigProtocol) {
+        self.providers = self.providersFor(config: config)
+        self.listeners = self.listenersFor(config: config)
```
saeedbashir, line 40: `self` is redundant.

OpenEdX/Managers/PushNotificationsManager/PushNotificationsManager.swift
```diff
@@ -0,0 +1,96 @@
...
+    }
+    
+    private func providersFor(config: ConfigProtocol) -> [PushNotificationsProvider] {
+        var rProviders: [PushNotificationsProvider] = []
```
saeedbashir, line 44: How about renaming it to providers, pushProviders or pProviders? as rProviders doesn't go with the feature.

OpenEdX/Managers/PushNotificationsManager/PushNotificationsManager.swift
```diff
@@ -0,0 +1,96 @@
...
+            rProviders.append(FCMProvider())
+        }
+        if config.braze.pushNotificationsEnabled {
+            rProviders.append(BrazeProvider())
+        }
+        return rProviders
+    }
+    
+    private func listenersFor(config: ConfigProtocol) -> [PushNotificationsListener] {
+        var rListeners: [PushNotificationsListener] = []
```
saeedbashir, line 55: How about renaming it to listeners, pushlisteners, or pListeners? as rListeners doesn't go with the feature.

OpenEdX/Managers/PushNotificationsManager/Listeners/BrazeListener.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  BrazeListener.swift
+//  OpenEdX
+//
+//  Created by Anton Yarmolenka on 24/01/2024.
+//
+
+import Foundation
+
+class BrazeListener: PushNotificationsListener {
```
saeedbashir, line 10: I guess it should also implement `didReceiveRemoteNotification`

saeedbashir (changes requested):

OpenEdX/Managers/DeepLinkManager/DeepLinkManager.swift
rnr: I will change comment. DeepLinkManager can get links from deep link service and also link can be extracted from push notifications body. So we have now two functions to process these links: 
`func processDeepLink(with params: [String: Any]) {`
and
`func processLinkFromNotification(_ link: PushLink) {`
both functions should try to detect redirection type and parameters and call redirection process

OpenEdX/Managers/PushNotificationsManager/PushNotificationsManager.swift
rnr: Braze and FCM notifications have different formats, and if we have two listeners enabled, when calling the `didReceiveRemoteNotification` function, we must determine that it is a notification from the correct service. Otherwise we will try to retrieve the data from the wrong format and it will fail 99% of the time.

OpenEdX/Managers/PushNotificationsManager/Listeners/FCMListener.swift
rnr: I added this function to the `PushNotificationsListener` protocol as an extension because it does the same thing for all listeners - checks this notification came from that listener, creates a `PushLink` from `userInfo` and calls the `processLinkFromNotification` function from `DeepLinkManager`.

OpenEdX/Managers/PushNotificationsManager/Listeners/BrazeListener.swift
rnr: https://github.com/openedx/openedx-app-ios/pull/263#discussion_r1482953608

forgotvas (dismissed):

OpenEdX/AppDelegate.swift
```diff
@@ -70,6 +75,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]
     ) -> Bool {
         if let config = Container.shared.resolve(ConfigProtocol.self) {
```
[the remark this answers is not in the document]
saeedbashir: Although it's not part of the underlying PR, but as we are, how about fixing it as well, like the below code for facebook can return false, whereas the method should have been called for Google or microsoft. It would be nice if we see if the URL is for facebook only then it will be returning the status.
```
if config.facebook.enabled {
            ApplicationDelegate.shared.application(
                app,
                open: url,
                sourceApplication: options[UIApplication.OpenURLOptionsKey.sourceApplication] as? String,
                annotation: options[UIApplication.OpenURLOptionsKey.annotation]
            )
        }
```

OpenEdX/AppDelegate.swift
saeedbashir: It would be nice if we can follow the same pattern for `didReceiveRemoteNotification` and `configureDeepLinkServices`.

OpenEdX/AppDelegate.swift
```diff
@@ -62,33 +63,44 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
-                )
-            }
+        if config.facebook.enabled {
+            ApplicationDelegate.shared.application(
+                app,
+                open: url,
+                sourceApplication: options[UIApplication.OpenURLOptionsKey.sourceApplication] as? String,
+                annotation: options[UIApplication.OpenURLOptionsKey.annotation]
+            )
+        }
```
saeedbashir, line 93: Although it's not part of the underlying PR, but as we are, how about fixing it as well, it can return false, whereas the method should have been called for Google or Microsoft. It would be nice if we see if the URL is for Facebook only then it will be returning the status.

OpenEdX/AppDelegate.swift
```diff
@@ -125,4 +137,30 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
+        guard let pushManager = Container.shared.resolve(PushNotificationsManager.self) else { return }
+        pushManager.didFailToRegisterForRemoteNotificationsWithError(error: error)
+    }
+    func application(
+        _ application: UIApplication,
+        didReceiveRemoteNotification userInfo: [AnyHashable: Any],
+        fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void
+    ) {
+        if let pushManager = Container.shared.resolve(PushNotificationsManager.self) {
```
saeedbashir, line 154: It would be nice if guard let can be used here as well.

OpenEdX/AppDelegate.swift
```diff
@@ -125,4 +137,30 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            pushManager.didReceiveRemoteNotification(userInfo: userInfo)
+        }
+        completionHandler(.newData)
+    }
+    
+    // Deep link
+    func configureDeepLinkServices(launchOptions: [UIApplication.LaunchOptionsKey: Any]?) {
+        if let deepLinkManager = Container.shared.resolve(DeepLinkManager.self) {
```
saeedbashir, line 162: It would be nice if guard let can be used here as well.

OpenEdX/Managers/DeepLinkManager/DeepLinkManager.swift
```diff
@@ -0,0 +1,72 @@
...
+    func configureWith(launchOptions: [UIApplication.LaunchOptionsKey: Any]?)
+    func handledURLWith(app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any]) -> Bool
+}
+
+class DeepLinkManager {
+    private var services: [DeepLinkService] = []
+    
+    // Init manager
+    public init(config: ConfigProtocol) {
+        services = servicesFor(config: config)
```
saeedbashir, line 22: How about renaming `servicesFor` to something like `configureServices(_ with config)` and call it. The method will populate `services` for all enabled services?

OpenEdX/Managers/DeepLinkManager/DeepLinkManager.swift
```diff
@@ -0,0 +1,72 @@
...
+    ) -> Bool {
+        for service in services where service.handledURLWith(app: app, open: url, options: options) {
+            return true
+        }
+        return false
+    }
+    
+    // This method do redirect with link from push notification
+    func processLinkFromNotification(_ link: PushLink) {
+        if anyServiceEnabled {
```
saeedbashir, line 60: I guess this check isn't valid because it's for deep link services not for the push notifications.

OpenEdX/Managers/PushNotificationsManager/PushNotificationsManager.swift
saeedbashir: It seems this method is to check whether the notification is for the listener or not. If that so who about renaming it to like canProcessPushNotification or something like that thoughts?

OpenEdX/Managers/DeepLinkManager/DeepLinkManager.swift
rnr: I don't agree with this name - this func doesn't configure services but just inits and returns services for current config

OpenEdX/Managers/PushNotificationsManager/PushNotificationsManager.swift
rnr: This method is not for checking `can or not` process this notification. It's for checking `should or not` this Listener handle with this notification at all (or this is for other Listener).
We're developing an abstraction layer, aren't? I believe we can rename any function during implementation if we see it does something other, no?

OpenEdX/Managers/DeepLinkManager/DeepLinkManager.swift
saeedbashir: ok

OpenEdX/Managers/DeepLinkManager/DeepLinkManager.swift
```diff
@@ -0,0 +1,70 @@
...
+class DeepLinkManager {
+    private var services: [DeepLinkService] = []
+    
+    // Init manager
+    public init(config: ConfigProtocol) {
+        services = servicesFor(config: config)
+    }
+    
+    private func servicesFor(config: ConfigProtocol) -> [DeepLinkService] {
+        var deepServices: [DeepLinkService] = []
```
saeedbashir, line 26: We are not going with method name renaming then how about renaming it to services?

saeedbashir (approved):
Just a small nit, otherwise LGTM 👍

OpenEdX/Managers/DeepLinkManager/DeepLinkManager.swift
rnr: It's just a `var` to return from a function. Why is its name so important if you are asking me to rename it a second time?

saeedbashir: I believe the names should be meaningful according to their usage. It doesn't matter whether they are local or instance variables.

saeedbashir: It can be renamed to services or deeplinkServices as per the usage.

rnr: Do you see a big difference between `deepServices` and `deeplinkServices`? Is it not "meaningful" enough?
Here https://github.com/openedx/openedx-app-ios/pull/263#discussion_r1481674064 you asked me to rename to 'pushProviders' - why not to 'pushnotificationProviders' to ' be meaningful according to their usage'? I'm just following your example.
 Don't you think we're just wasting time here?

rnr: I believe var `deepServices` of type `[DeepLinkService]` is enough 'readable' (especially since it's not used anywhere but these 5 lines of code)

saeedbashir: hmm ok if you are reluctant to make this change, then lets go with it.

saeedbashir (approved):

volodymyr-chekyrta (approved):

Issue #244, opened 2024-01-19
[iOS] Video continues to play when another video starts


sergeymomot:
Open any sequence with several video modules and start playing a first video, then choose another module and start playing second video (this step can be repeated many times). Result: first video continues to play when second (third, fourth.....) video(s) start playing.

https://github.com/openedx/openedx-app-ios/assets/2964930/d76dcad3-db08-42e0-82e3-4384cc48f686

PR #268, opened 2024-02-01
Fix/memory leak

forgotvas:
regarding to bug [iOS] Video continues to play when another video starts #244


forgotvas:
wrong target branch

PR #269, opened 2024-02-01
Fix/memory leak

forgotvas:
### This PR contains fix for:

* https://github.com/openedx/openedx-app-ios/issues/244

>Open any sequence with several video modules and start playing a first video, then choose another module and start playing second video (this step can be repeated many times). Result: first video continues to play when second (third, fourth.....) video(s) start playing.

https://github.com/openedx/openedx-app-ios/assets/2964930/d76dcad3-db08-42e0-82e3-4384cc48f686


rnr (dismissed):

Course/Course/Presentation/Video/PlayerViewController.swift
```diff
@@ -52,40 +49,66 @@ struct PlayerViewController: UIViewControllerRepresentable {
...
-            let currentSeconds = CMTimeGetSeconds(time)
-            guard let duration = controller.player?.currentItem?.duration else { return }
-            let totalSeconds = CMTimeGetSeconds(duration)
-            progress = Float(currentSeconds / totalSeconds)
-            currentProgress(progress, currentSeconds)
-        }
-    }
-    
     func updateUIViewController(_ playerController: AVPlayerViewController, context: Context) {
         DispatchQueue.main.async {
```
saeedbashir, line 53: @forgotvas As we are here, how about defining `updateUIViewController` as `@MainActor`?

forgotvas: @saeedbashir to be honestly i don't think that `DispatchQueue.main.async` required here, because updateUIViewController it's a swiftUI view life cycle method that should guarantee that it will be called from main thread. I will try to test that without dispatch and will commit changes.

forgotvas: @saeedbashir have removed `DispatchQueue.main.async` and tested with enabled `Main thread checker` (enabled by default) and `Thread sanitizer` diagnostics tools - no problem there.

rnr (approved):

volodymyr-chekyrta (approved):

saeedbashir (approved):

Issue #240, opened 2024-01-17
[iOS] Dark Mode Support for the webview


sergeymomot:
Add dark mode support for the webview

<img width="280" alt="043d9876-cd47-4182-bf7e-1e0098ae6498" src="https://github.com/openedx/openedx-app-ios/assets/2964930/043d9876-cd47-4182-bf7e-1e0098ae6498">

PR #274, opened 2024-02-05
Feat/dark mode webview

forgotvas:
### PR contains fix for:

- #240

Simulator Screenshot - iPhone 15 - 2024-02-05 at 11 18 42
Simulator Screenshot - iPhone 15 - 2024-02-05 at 11 18 45


rnr (approved):

volodymyr-chekyrta (approved):
Looks good to me 👍

Core/Core/View/Base/Webview/Models/ColorInvertionInjection.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  ColorInvertionInjection.swift
+//  Core
+//
+//  Created by Vadim Kuznetsov on 31.01.24.
+//
+
+import WebKit
+
+public struct ColorInvertionInjection: WebViewScriptInjectionProtocol, CSSInjectionProtocol {
```
miankhalid, line 10: minor nit: Invertion isn't a valid word, I'd suggest renaming the class to `ColorInversionInjection`

forgotvas: renamed

rnr (approved):

volodymyr-chekyrta (approved):

Core/Core/View/Base/Webview/WebView.swift
```diff
@@ -120,7 +61,24 @@ public struct WebView: UIViewRepresentable {
             addObserver()
         }
 
+        public func webView(_ webView: WKWebView, didStartProvisionalNavigation navigation: WKNavigation!) {
+            webView.isHidden = true
+        }
```
saeedbashir, line 66: Is it important to hide the webview here? as hiding it here is required `webview?isHidded = true` at multiple places

Core/Core/View/Base/Webview/WebView.swift
```diff
@@ -120,7 +61,24 @@ public struct WebView: UIViewRepresentable {
...
+        
+        public func webView(_ webView: WKWebView, didFail navigation: WKNavigation!, withError error: Error) {
+            webView.isHidden = false
```
saeedbashir, line 69: We need to create a task for handling the review failure case.  In which we might add a loading error with the reload button. Thoughts?

Course/Course/Presentation/Unit/CourseUnitViewModel.swift
```diff
@@ -16,11 +16,12 @@ public enum LessonType: Equatable {
     case discussion(String, String, String)
     
     static func from(_ block: CourseBlock, streamingQuality: StreamingQuality) -> Self {
+        let mandatoryInjections: [WebviewInjection] = [.inversionCss, .ajaxCallback]
```
saeedbashir, line 19: The name is fine, but how about renaming it to `webviewInjections` thoughts?

Core/Core/View/Base/Webview/Models/WebviewInjection.swift
```diff
@@ -47,6 +47,11 @@ public extension WebviewInjection {
             .webviewInjection()
     }
 
+    static var inversionCss: WebviewInjection {
```
saeedbashir, line 50: @forgotvas Although the name is fine, but what do you think about renaming it to colorInversionCss, for more clear name as inversion can be of any type thoughts?

Core/Core/View/Base/Webview/WebView.swift
forgotvas: @rnr ^^^

Course/Course/Presentation/Unit/CourseUnitViewModel.swift
forgotvas: i'm not sure that it will be correct, idea is to name that variable to indicate that these are the necessary injections for all webviews.

Core/Core/View/Base/Webview/Models/WebviewInjection.swift
forgotvas: renamed

Core/Core/View/Base/Webview/WebView.swift
saeedbashir: Is it possible to use `isLoading` property here, so the user will see loading while the inversion is happening?

forgotvas: no, `isLoading` property doesn't hide webView it's shows `ProgressView` on parent view.

Core/Core/View/Base/Webview/WebView.swift
rnr: I agree it might be separate ticket

Core/Core/View/Base/Webview/WebView.swift
saeedbashir: Actually, that's what I'm trying to understand why it's needed, because if it's user-facing and there won't be any loading screen, and we also hide the webview, the view will not look good. If we se the isloading true, the user will see a loading screen and behind the scene the color inversion will happen.

forgotvas: just to clarify, do you want to set `isHidden` when we set `isLoading`? If so then it will work.

forgotvas: without `isHidden`:

https://github.com/openedx/openedx-app-ios/assets/480059/49f436f6-2f72-4608-8b16-44ac480d2e70

with `isHidden`

https://github.com/openedx/openedx-app-ios/assets/480059/7302a350-4e0e-44ab-8dff-30a6d4b0a161

saeedbashir (approved):
LGTM 👍

Issue #275, opened 2024-02-05
[iOS] Data for Programs tab isn't displayed


sergeymomot:
1. Install app by deleting the previous version
2. Log in
3. Navigate to Programs tab: only infinite loading indicator is shown
4. Navigate to Dashboard -> open any available Course -> open any content
5. Again go to Programs tab: now the contents of the Programs tab will be displayed normally

Content of the Programs tab should be displayed, without any additional actions.

https://github.com/openedx/openedx-app-ios/assets/2964930/b507ae49-5bc2-4f8d-a32b-3a897a3c559d

PR #276, opened 2024-02-06
chore: cookies refresh in enrolled programs webview

saeedbashir:
This PR adds functionality of refresh cookies to programs webview.

[Data for Programs tab isn't displayed](https://github.com/openedx/openedx-app-ios/issues/275)

### DIFF
--- a/Core/Core/View/Base/WebUnitViewModel.swift
+++ b/Core/Core/View/Base/WebUnitViewModel.swift
@@ -8,17 +8,17 @@
 import Foundation
 import SwiftUI
 
-public class WebUnitViewModel: ObservableObject {
+public class WebUnitViewModel: ObservableObject, WebviewCookiesUpdateProtocol {
     
-    let authInteractor: AuthInteractorProtocol
+    public let authInteractor: AuthInteractorProtocol
     let config: ConfigProtocol
     
-    @Published var updatingCookies: Bool = false
-    @Published var cookiesReady: Bool = false
-    @Published var showError: Bool = false
+    @Published public var updatingCookies: Bool = false
+    @Published public var cookiesReady: Bool = false
+    @Published public var showError: Bool = false
     private var retryCount = 1
     
-    var errorMessage: String? {
+    public var errorMessage: String? {
         didSet {
             withAnimation {
                 showError = errorMessage != nil
@@ -29,27 +29,5 @@ public class WebUnitViewModel: ObservableObject {
     public init(authInteractor: AuthInteractorProtocol, config: ConfigProtocol) {
         self.authInteractor = authInteractor
         self.config = config
-    }
-    
-    @MainActor
-    func updateCookies(force: Bool = false) async {
-        guard !updatingCookies else { return }
-        do {
-            updatingCookies = true
-            try await authInteractor.getCookies(force: force)
-            cookiesReady = true
-            updatingCookies = false
-            errorMessage = nil
-        } catch {
-            if error.isInternetError {
-                errorMessage = CoreLocalization.Error.slowOrNoInternetConnection
-            } else if retryCount > 0 {
-                retryCount -= 1
-                await updateCookies(force: force)
-            } else {
-                errorMessage = CoreLocalization.Error.unknownError
-            }
-            updatingCookies = false
-        }
     }
 }
--- a/Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewView.swift
+++ b/Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewView.swift
@@ -61,7 +61,11 @@ public struct ProgramWebviewView: View {
                         baseURL: ""
                     ),
                     isLoading: $isLoading,
-                    refreshCookies: {},
+                    refreshCookies: {
+                        await viewModel.updateCookies(
+                            force: true
+                        )
+                    },
                     navigationDelegate: viewModel
                 )
                 
--- a/Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewView.swift
+++ b/Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewView.swift
@@ -69,7 +69,7 @@ public struct ProgramWebviewView: View {
                     navigationDelegate: viewModel
                 )
                 
-                if isLoading || viewModel.showProgress {
+                if isLoading || viewModel.showProgress || viewModel.updatingCookies {
                     HStack(alignment: .center) {
                         ProgressBar(
                             size: 40,
@@ -92,6 +92,7 @@ public struct ProgramWebviewView: View {
                             viewModel.errorMessage = nil
                         }
                     }
+                    .animation(.default, value: 1)
                 }
             }
             
--- a/Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewView.swift
+++ b/Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewView.swift
@@ -92,7 +92,6 @@ public struct ProgramWebviewView: View {
                             viewModel.errorMessage = nil
                         }
                     }
-                    .animation(.default, value: 1)
                 }
             }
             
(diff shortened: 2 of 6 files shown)


Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewViewModel.swift
```diff
@@ -14,6 +14,9 @@ public class ProgramWebviewViewModel: ObservableObject {
     @Published var courseDetails: CourseDetails?
     @Published private(set) var showProgress = false
     @Published var showError: Bool = false
+    @Published var updatingCookies: Bool = false
```
forgotvas, line 17: hi @saeedbashir, where is published property `updatingCookies` using?

Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewViewModel.swift
```diff
@@ -14,6 +14,9 @@ public class ProgramWebviewViewModel: ObservableObject {
...
+    private var retryCount = 1
+    
     var errorMessage: String? {
         didSet {
             withAnimation {
```
forgotvas, line 22: it's bad practice to use `withAnimation` block inside view model, better to use animation modifier in that case, something like `.animation(.linear(duration: 0.25), value: showError)`

Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewViewModel.swift
```diff
@@ -94,6 +100,27 @@ public class ProgramWebviewViewModel: ObservableObject {
             return .enrollOpen
         }
     }
+    
+    @MainActor
+    func updateCookies(force: Bool = false) async {
```
forgotvas, line 105: it's same method as in WebUnitViewModel better to move it to one place to be reusable if it will need in other place. Protocol for example.

forgotvas (changes requested):

Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewViewModel.swift
saeedbashir: @forgotvas Underneath, they are using the shared code, which is `authInteractor.getCookies(force: force)` and this method is just for handling the errors etc, do we still need to move this method to a common place?

Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewViewModel.swift
saeedbashir: This is what we are using almost everywhere in the app. We might create a task to update it for the whole app. What do you say?

forgotvas: @sergeymomot can you create ticket about that? But as we are here, @saeedbashir could you change it in that view?

Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewViewModel.swift
saeedbashir: So this is what you meant when you said to move it to common place, or you have something else on mind?

```
public protocol UpdateCookies {
    var updatingCookies: Bool { get }
    var errorMessage: String? { get }
    var retryCount: Int { get }
    func updateCookies(force: Bool) async
}

public class UpdateCookiesModel: UpdateCookies {
    @Published public var updatingCookies: Bool
    @State public var errorMessage: String?
    public var retryCount: Int
    let authInteractor: AuthInteractorProtocol
    
    init(updatingCookies: Bool, errorMessage: String?, retryCount: Int, authInteractor: AuthInteractorProtocol) {
        self.updatingCookies = updatingCookies
        self.errorMessage = errorMessage
        self.authInteractor = authInteractor
        self.retryCount = retryCount
    }
    
    public func updateCookies(force: Bool) async {
        guard !updatingCookies else { return }
        do {
            updatingCookies = true
            try await authInteractor.getCookies(force: force)
            updatingCookies = false
            errorMessage = nil
        } catch {
            if error.isInternetError {
                errorMessage = CoreLocalization.Error.slowOrNoInternetConnection
            } else if retryCount > 0 {
                retryCount -= 1
                await updateCookies(force: force)
            } else {
                errorMessage = CoreLocalization.Error.unknownError
            }
            updatingCookies = false
        }
    }
}

Comment (forgotvas) [reply Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewViewModel.swift:105]: No, in my mind i see 2 options:
1) **Protocols**. Protcol that contains function and it doesn't to be described twice, only required variables: https://gist.github.com/forgotvas/ad4980a16731127fb7ddde37746000eb
2) **Inheritance**. In that case we are using base ObservableObject that contains properties and method: https://gist.github.com/forgotvas/c4330000d3dc22de514fec1c64543d4a

File: Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewView.swift [hunk 4]
```diff
@@ -61,11 +61,15 @@ public struct ProgramWebviewView: View {
...
+                    refreshCookies: {
+                        await viewModel.updateCookies(
+                            force: true
+                        )
+                    },
                     navigationDelegate: viewModel
                 )
                 
-                if isLoading || viewModel.showProgress {
+                if isLoading || viewModel.showProgress || viewModel.updatingCookies {
```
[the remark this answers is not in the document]
forgotvas: will be something like that:

https://github.com/openedx/openedx-app-ios/assets/480059/2f328a50-b62f-4057-b930-2de218b1482f

Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewViewModel.swift
sergeymomot: > @sergeymomot can you create ticket about that? But as we are here, @saeedbashir could you change it in that view?

https://github.com/openedx/openedx-app-ios/issues/277

Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewView.swift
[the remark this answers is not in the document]
saeedbashir: > @rnr maybe we should create ticket there?

This is on radar as the error handling isn't done the way we want; at most places, it's auto dismissible snackbar which isn't a great way to inform users about complete failures.

Discovery/Discovery/Presentation/WebPrograms/ProgramWebviewView.swift
```diff
@@ -88,6 +92,7 @@ public struct ProgramWebviewView: View {
                             viewModel.errorMessage = nil
                         }
                     }
+                    .animation(.default, value: 1)
```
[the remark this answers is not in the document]
saeedbashir: As you are able to see errors, do I need to move the `.animation` to parent view?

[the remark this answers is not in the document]
saeedbashir: The animation is smooth with parent view so moved the `.animation` modifier to parent view.

Core/Core/View/Base/WebUnitViewModel.swift
```diff
@@ -8,17 +8,17 @@
...
+    public let authInteractor: AuthInteractorProtocol
     let config: ConfigProtocol
     
-    @Published var updatingCookies: Bool = false
-    @Published var cookiesReady: Bool = false
-    @Published var showError: Bool = false
+    @Published public var updatingCookies: Bool = false
+    @Published public var cookiesReady: Bool = false
+    @Published public var showError: Bool = false
     private var retryCount = 1
```
forgotvas, line 19: property `retryCount` was defined and never used same inside `ProgramWebviewViewModel`

forgotvas (changes requested):

forgotvas (approved):
looks good for me

rnr (approved):

volodymyr-chekyrta (approved):

Issue #179, opened 2023-11-27
[iOS] Assessment content


nizmaylova:
Add Assessment content support  in the mobile app.

[Link to content](https://learning-raccoonapis.sandbox.edx.org/course/course-v1:edx+DW101+2023_DW/block-v1:edx+DW101+2023_DW+type@sequential+block@1e5cd9f233a2453f83731ccbd863b731/block-v1:edx+DW101+2023_DW+type@vertical+block@f467da80028f4f83b35707e873edfe88)

moiz994:
Implemented through webview in the Discover tab. Closing ticket.

Issue #175, opened 2023-11-27
[iOS] Peer Instruction Tool content


nizmaylova:
Add Peer Instruction Tool content support in the mobile app.

[Link to content](https://learning-raccoonapis.sandbox.edx.org/course/course-v1:OrgX+Demo101+2023_Q2/block-v1:OrgX+Demo101+2023_Q2+type@sequential+block@52e9412a5700455b9e88948c5e61fd6f/block-v1:OrgX+Demo101+2023_Q2+type@vertical+block@9532b780fcc4443780ebe15c28806c48)

![Image](https://github.com/openedx/openedx-app-ios/assets/121940663/9c617872-be07-443a-935d-3eee3f858a9f)
Note: Current edX app from store is shown in the attachment

PR #282, opened 2024-02-13
Open assessment and Peer instruction tool contents types

forgotvas:
### That PR contains:

- #175
- #179


rnr (dismissed):

saeedbashir (commented):
@forgotvas On the [edX codebase](https://github.com/openedx/edx-app-ios) we haven't explicitly whitelisted xblock type `ubcpi` and it's being rendered in the app because of an [additional condition](https://github.com/openedx/edx-app-ios/blob/master/Source/OEXRouter%2BSwift.swift#L47). I am unsure how many xblocks will be rendered because of the additional condition. It would be nice if we would mimic the same logic here. The following doing the magic in the edX codebase.
```
case .Unknown(_), .HTML: return multiDevice ? .HTML(.Base) : .Unknown
```

volodymyr-chekyrta (dismissed):

forgotvas:
> @forgotvas On the [edX codebase](https://github.com/openedx/edx-app-ios) we haven't explicitly whitelisted xblock type `ubcpi` and it's being rendered in the app because of an [additional condition](https://github.com/openedx/edx-app-ios/blob/master/Source/OEXRouter%2BSwift.swift#L47). I am unsure how many xblocks will be rendered because of the additional condition. It would be nice if we would mimic the same logic here. The following doing the magic in the edX codebase.
> 
> ```
> case .Unknown(_), .HTML: return multiDevice ? .HTML(.Base) : .Unknown
> ```

done

saeedbashir (dismissed):
LGTM 👍

rnr (approved):

volodymyr-chekyrta (approved):

PR #289, opened 2024-02-20
feat: Don't delete downloaded videos when the user logs out

eyatsenkoperpetio:
When a user logs out, the downloaded content should not be deleted, so that when the user logs in again, all downloaded content will be available to him again.

For each user, only their downloaded content should be displayed.

### DIFF
--- a/Profile/Profile/Data/ProfileRepository.swift
+++ b/Profile/Profile/Data/ProfileRepository.swift
@@ -71,8 +71,6 @@ public class ProfileRepository: ProfileRepositoryProtocol {
             ProfileEndpoint.logOut(refreshToken: refreshToken, clientID: config.oAuthClientId)
         )
         storage.clear()
-        //await downloadManager.deleteAllFiles()
-        //coreDataHandler.clear()
     }
     
     public func getSpokenLanguages() -> [PickerFields.Option] {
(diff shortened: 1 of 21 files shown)


Profile/Profile/Data/ProfileRepository.swift
```diff
@@ -71,8 +71,8 @@ public class ProfileRepository: ProfileRepositoryProtocol {
             ProfileEndpoint.logOut(refreshToken: refreshToken, clientID: config.oAuthClientId)
         )
         storage.clear()
-        await downloadManager.deleteAllFiles()
-        coreDataHandler.clear()
+        //await downloadManager.deleteAllFiles()
```
rnr, line 74: do we need these commented out lines?

rnr (dismissed):

Profile/Profile/Data/ProfileRepository.swift
```diff
@@ -71,8 +71,8 @@ public class ProfileRepository: ProfileRepositoryProtocol {
...
+        //coreDataHandler.clear()
```
saeedbashir, line 75: @eyatsenkoperpetio Do we need to keep this as well if we want to keep downloaded files?

forgotvas (approved):

rnr (approved):

saeedbashir (approved):
LGTM 👍

volodymyr-chekyrta (approved):

PR #291, opened 2024-02-21
Firebase and Segment integration

rnr:
This PR implements both Analytics options - `Firebase+GoogleAnalytics` standalone library and `Segment+GoogleAnalytics`.
In the config we have the `ANALYTICS_SOURCE` parameter:
```yaml
FIREBASE:
    ENABLED: true
    ANALYTICS_SOURCE: 'segment'
```
This parameter may have values:
- firebase
- segment

Depending on this parameter, the application uses the appropriate library. For this purpose, `FirebaseAnalyticsService` and `SegmentAnalyticsService` are added. Both services are added to the DI.


Core/Core/Configuration/Config/SegmentConfig.swift
```diff
@@ -0,0 +1,31 @@
...
+}
+
+public final class SegmentConfig: NSObject {
+    public var enabled: Bool = false
+    public var writeKey: String = ""
+    
+    init(dictionary: [String: AnyObject]) {
+        super.init()
+        enabled = dictionary[SegmentKeys.enabled] as? Bool == true
+        writeKey = dictionary[SegmentKeys.writeKey] as? String ?? ""
```
saeedbashir, line 22: Let's make `enabled` also dependent on `writeKey` like the writeKey shouldn't be empty.

Authorization/Authorization/Presentation/Registration/SignUpViewModel.swift
```diff
@@ -131,7 +131,7 @@ public class SignUpViewModel: ObservableObject {
                 fields: validateFields,
                 isSocial: externalToken != nil
             )
-            analytics.setUserID("\(user.id)")
+            analytics.identify(id: "\(user.id)", username: user.username, email: user.email)
```
saeedbashir, line 134: How about moving the `analytics.identify(.....)` call to `showMainOrWhatsNewScreen` so you don't have to worry about from where the main screen is being shown?  At the moment it's being called from multiple places.

OpenEdX/Managers/FirebaseAnalyticsManager/FirebaseAnalyticsManager.swift
```diff
@@ -0,0 +1,28 @@
...
+//  OpenEdX
+//
+//  Created by Anton Yarmolenka on 19/02/2024.
+//
+
+import Foundation
+import Firebase
+import Core
+
+class FirebaseAnalyticsManager: AnalyticsService {
```
saeedbashir, line 12: Although we have collaborated on it and name is also fine, but as it's more like a tracker instead of manager, how about renaming it to `FirebaseAnalyticsTracker`.

If it sounds good then the same will go for Segment.

OpenEdX/Managers/SegmentAnalyticsManager/SegmentAnalyticsManager.swift
```diff
@@ -0,0 +1,44 @@
...
+    
+    // Init manager
+    public init(config: ConfigProtocol) {
+        guard config.segment.enabled else { return }
+
+        let configuration = Configuration(writeKey: config.segment.writeKey)
+                        .trackApplicationLifecycleEvents(true)
+                        .flushInterval(10)
+        analytics = Analytics(configuration: configuration)
+        if config.firebase.isAnalyticsSourceSegment {
```
saeedbashir, line 24: Oh it got missed, please update condition for `config.firebase.enabled` as well so in case of missing keys app will not crash.

OpenEdX/Managers/FirebaseAnalyticsManager/FirebaseAnalyticsManager.swift
rnr: What about `FirebaseAnalyticsService`?

Authorization/Authorization/Presentation/Registration/SignUpViewModel.swift
rnr: I don't really like the idea of passing User variable inside the router. However, we may come back to this question when we work with analytics (this is not part of this PR)

Core/Core/Configuration/Config/SegmentConfig.swift
rnr: done

OpenEdX/Managers/FirebaseAnalyticsManager/FirebaseAnalyticsManager.swift
rnr: done

OpenEdX/Managers/SegmentAnalyticsManager/SegmentAnalyticsManager.swift
rnr: done

forgotvas (approved):

saeedbashir (approved):
LGTM 👍

Authorization/Authorization/Presentation/Registration/SignUpViewModel.swift
saeedbashir: ok

OpenEdX/AppDelegate.swift
```diff
@@ -35,13 +32,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
     ) -> Bool {
         initDI()
-        
         if let config = Container.shared.resolve(ConfigProtocol.self) {
             Theme.Shapes.isRoundedCorners = config.theme.isRoundedCorners
-            if let configuration = config.firebase.firebaseOptions {
-                FirebaseApp.configure(options: configuration)
-                Crashlytics.crashlytics().setCrashlyticsCollectionEnabled(true)
-            }
```
volodymyr-chekyrta, line 44: We can't remove that part of the code, Crashlytics will stop working.

OpenEdX/Managers/FirebaseAnalyticsService/FirebaseAnalyticsService.swift
```diff
@@ -0,0 +1,28 @@
...
+import Foundation
+import Firebase
+import Core
+
+class FirebaseAnalyticsService: AnalyticsService {
+    // Init manager
+    public init(config: ConfigProtocol) {
+        guard config.firebase.enabled && config.firebase.isAnalyticsSourceFirebase else { return }
+
+        FirebaseApp.configure()
```
volodymyr-chekyrta, line 17: Since we don't have the GoogleServices.plist file to configure Firebase, we have to pass the configuration to the configure function.
https://github.com/openedx/openedx-app-ios/blob/main/OpenEdX/AppDelegate.swift#L42

OpenEdX/AppDelegate.swift
saeedbashir: @volodymyr-chekyrta We've added [FirebaseCrashlyticsCollectionEnabled ](https://github.com/openedx/openedx-app-ios/pull/291/files#diff-e70052e8a1bf7ccec67056e9557328a349fe3aa78447addbc7d65542d6515c08R32) in the info.plist, I believe this will do the job.

OpenEdX/Managers/FirebaseAnalyticsService/FirebaseAnalyticsService.swift
saeedbashir: @volodymyr-chekyrta  The [config processing script](https://github.com/openedx/openedx-app-ios/blob/main/config_script/process_config.py#L156) is already generating the `GoogleService-Info.plist` file by reading the values from the config for `FIREBASE`.

volodymyr-chekyrta: I see, thank you!

volodymyr-chekyrta (approved):
@rnr, LGTM 👍 
@saeedbashir, thanks for comments

forgotvas (approved):

PR #311, opened 2024-03-01
feat: use custom font and colors across the app where were missing

saeedbashir:
At a number of places, the default font was being used for so many elements, and the same was the case for colors. At a number of place, like the `.red` color was used at a number of places instead of the custom `Alert` color. 

This PR fixes the issue and enhances the app theme capability by applying the app defined colors and app defined custom fonts across the app. 

Note: 

1. This PR will not impact the openedX default theme. The colors scheme will remain the same. 
2. The default font is still being used for picker (discussion filters, discussion sort), as there isn't a straight-forward way to use custom fonts for picker items. Will look into it later.


rnr (dismissed):

volodymyr-chekyrta (dismissed):

rnr (dismissed):
looks good to me
please fix one issue for edx dark mode 
Screenshot 2024-03-06 at 13 39 26

rnr:
I also see some font size changes for openEdX on the login page. Please confirm if this is acceptable. 
Screenshot 2024-03-06 at 14 52 11

saeedbashir:
@rnr feedback addressed and I've corrected the size of sign-in screen elements

rnr:
> looks good to me please fix one issue for edx dark mode Screenshot 2024-03-06 at 13 39 26

@saeedbashir please confirm we don't need border for Cancel button here in dark mode

rnr (approved):

saeedbashir:
> > looks good to me please fix one issue for edx dark mode Screenshot 2024-03-06 at 13 39 26
> 
> @saeedbashir please confirm we don't need border for Cancel button here in dark mode

The screenshot is not accessible.

OpenEdX/Info.plist
```diff
@@ -27,9 +29,16 @@
...
 	<array/>
+	<key>UILaunchScreen</key>
+	<dict>
+		<key>UIColorName</key>
+		<string>SplashBackground</string>
+		<key>UIImageName</key>
+		<string>appLogo</string>
+		<key>UIImageRespectsSafeAreaInsets</key>
+		<true/>
+	</dict>
```
volodymyr-chekyrta, line 40: Is there any chance not to replace the `LaunchScreen.storyboard` with the `Info.plist` configuration?
For some OeX instances we need a slightly more customizable LaunchScreen than just a background color and image, and this would be much easier with a storyboard file.

volodymyr-chekyrta: Maybe the white-label script can just remove the `Launch screen interface file base name` key from the `Info.plist` and add `UILaunchScreen`?

volodymyr-chekyrta: Replacing the whole LaunchScreen.storyboard also an option 👍

rnr: @volodymyr-chekyrta Yes, we can try both approaches. At first glance, replacing the `LaunchScreen.storyboard` file looks easier. @saeedbashir could you please prepare and add this file into `edx-mobile-config` repo  and I will add whitelabel config and script changes?
Thank you

volodymyr-chekyrta: Thanks!

saeedbashir: @volodymyr-chekyrta It's re-added.

PR #330, opened 2024-03-08
chore: configure branch via info.plist

saeedbashir:
In the [production codebase](https://github.com/openedx/dx-app-ios) we were using `Branch.setBranchKey(key)` to set the branch key and it was working as expected.

But with the new version of the branch SDK `3.x.x`, it seems it's no longer working and requires the keys to be added in the info.plist.

This PR fixes it and now the branch keys are being added to the info.plist via config_process script on compile time.

### Config
```
BRANCH:
    ENABLED: true
    KEY: 'key_live'
   `KEY_TEST`: 'key_test' # if this be missing KEY will be used as live and test keys and it works.
 
   ```


volodymyr-chekyrta (approved):
Thank you @saeedbashir @rnr 👍

rnr (approved):

Issue #359, opened 2024-03-20
[iOS] JSON parsing error is showing up in log when openeing the Dasbhoard viewmodel


OmarIthawi:
### Steps to reproduce

1. Where are you starting? What can you see?

A newly compiled app from `develop` branch connected to my server `blue.zeitlabs.com` which is Open edX `master` backend (more recent the Quince).

3. What do you click?

The dashboard button.

We got this error in the mobile error:

```
valueNotFound(Swift.Array<Swift.String>, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "enrollments", intValue: nil), CodingKeys(stringValue: "results", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "course", intValue: nil), CodingKeys(stringValue: "discussion_url", intValue: nil)], debugDescription: "Cannot get unkeyed decoding container -- found null value instead", underlyingError: nil))
```

### HTTP Request
```
Request: GET https://blue.zeitlabs.com/api/mobile/v3/users/omarstudent/course_enrollments?page=1


```
![image](https://github.com/openedx/openedx-app-ios/assets/645156/76ecdbe8-44e7-44dc-bed5-44e74bce8ddb)

#### What did you expect?
See list of courses.

#### What happened instead?
I see empty list dashboard.

### Your phone model

iPhone Simulator

### Operating system version

iPhone 15 Pro simulator - iOS 17.2

### Application version

develop branch (v1.0.0)

gurhub:
Hey @volodymyr-chekyrta, first of all, thanks for the solution and merge. After your solution, I manually got the raw Data_Dashboard.swift file content and tried it on my machine and I received another parsing error on Xcode's console:

`CodingKeys(stringValue: "start_display", intValue: nil)], debugDescription: "Cannot get unkeyed decoding container -- found null value instead", underlyingError: nil))`

By following your method, I was able to remove the start_display key from the **Data_Dashboard.swift** file. And run and test again. Now everything is working as expected. 

Best

PR #360, opened 2024-03-20
fix: Dashboard parsing error

volodymyr-chekyrta:
**Issue:** https://github.com/openedx/openedx-app-ios/issues/359

**Solution Description:**
Removed unused fields in the `Data_Dashboard.swift` to fix the parsing error.


openedx-webhooks:
Thanks for the pull request, @volodymyr-chekyrta! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

Core/Core/Data/Model/Data_Dashboard.swift
```diff
@@ -134,9 +131,6 @@ public extension DataLayer {
             case courseImage = "course_image"
             case courseAbout = "course_about"
             case courseSharingUtmParameters = "course_sharing_utm_parameters"
-            case courseUpdates = "course_updates"
-            case courseHandouts = "course_handouts"
-            case discussionURL = "discussion_url"
```
saeedbashir, line 139: @miankhalid in the production app, we are using `course_handouts` to show/add the Handouts in the course menu and likewise, we are using `discussion_url` to enable/disable course discussions and `course_updates` being used for course announcements. e.g if `discussion_url` is nil, discussion tab won't be added in the course menu on course dashboard screen.

Should we remove these or keep them and create a ticket to get parity with the production app?

Core/Core/Data/Model/Data_Dashboard.swift
```diff
@@ -112,9 +111,6 @@ public extension DataLayer {
         public let courseImage: String
         public let courseAbout: String
         public let courseSharingUtmParameters: CourseSharingUtmParameters
-        public let courseUpdates: String
-        public let courseHandouts: String
-        public let discussionURL: String
```
shafqat-muneer, line 117: Consider this idea: If we render all fields optional, it would prevent any disruptions in functionality, even if certain fields remain unused or no data is received from the backend.

shafqat-muneer (changes requested):

Core/Core/Data/Model/Data_Dashboard.swift
miankhalid: @volodymyr-chekyrta the only downside I'm seeing with this approach is we're fattening the response of the said API. Currently, there are courses on our side whose response for Blocks has reached size close to 2 to 3Mbs.

If we keep on adding stuff, it'll only increase further. Plus, Glib will have to update his PR https://github.com/openedx/edx-platform/pull/34273# even further.

For now, I do think we can update this API but in the longer run we should look into splitting these responses into atleast 2 APIs where:

1. one would just return Blocks
2. and the other one would return all the course related details

thoughts?

volodymyr-chekyrta: Sure, sounds reasonable 💯 👍! 
But I'd like to check out the size and speed difference before we make any decisions.

miankhalid: yep that's why I've asked if Glib can do some performance & load testing before and after the change in his PR.

shafqat-muneer (approved):

openedx-webhooks:
@volodymyr-chekyrta 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-app-ios&entry.867055334=https://github.com/openedx/openedx-app-ios/pull/360&entry.1484655318=https://github.com/volodymyr-chekyrta&entry.752974735=2024-03-20+15:41&entry.1917517419=2024-03-22+09:11&entry.2133058324=Yes) so we can improve your experience in the future.

PR #386, opened 2024-04-04
feat: Calendar Sync Feature Analytics Implementation

shafqat-muneer:
[LEARNER-9888](https://2u-internal.atlassian.net/browse/LEARNER-9888): Calendar Sync Feature Analytics Implementation

**Note:**
These analytics additionally necessitate the `EnrollmentMode` which is currently absent in "course_info/blocks" API. Once this information becomes available, we will address it separately. Currently, I have included this parameter in the Analytics events, but it is being passed as an empty string.


rnr (dismissed):

Core/Core/Analytics/CoreAnalytics.swift
```diff
@@ -244,3 +254,41 @@ public struct EventCategory {
...
+    case add = "Add"
+    case remove = "Remove"
+    case update = "Update"
+    case confirmed = "Confirmed"
+}
+
+public enum Snackbar: String {
+    case add = "Add"
+    case remove = "Remove"
+    case update = "Update"
```
saeedbashir, line 293: It would be nice if you could move this to Course Analytics and also keep all the actions, and stuff in snake case instead of camel case.

shafqat-muneer:
@saeedbashir Feedback has been addressed. Prepared another review. 🎉

rnr (approved):

volodymyr-chekyrta (approved):

saeedbashir (approved):

Issue #358, opened 2024-03-19
[iOS] Follow/Vote/Report marks disappears after refreshing Discussion page


sergeymomot:
Create or open any discussion post
then mark Vote/Follow/Report (all at once or separately, it doesn’t matter)
refresh the page
observe that mark(s) has disappeared

+ similar behavior also for responses

https://github.com/openedx/openedx-app-ios/assets/2964930/df97c748-8b64-46ae-a8c0-cb3a3a1d4df1

PR #393, opened 2024-04-11
fix for discussion pull to refresh

rnr:
This PR fixes issue https://github.com/openedx/openedx-app-ios/issues/358 when after pull-to-refresh user sees 'old' data.
Added actual data retrieval for "parent message" in Thread and Responses Views

Issue recording:

https://github.com/openedx/openedx-app-ios/assets/2964930/df97c748-8b64-46ae-a8c0-cb3a3a1d4df1


forgotvas (approved):

Discussion/Discussion/Presentation/Comments/Responses/ResponsesViewModel.swift
```diff
@@ -88,19 +90,23 @@ public class ResponsesViewModel: BaseResponsesViewModel, ObservableObject {
     }
     
     @MainActor
-    func getComments(commentID: String, parentComment: Post, page: Int) async -> Bool {
+    func getResponsesData(commentID: String, parentComment: Post, page: Int, refresh: Bool = false) async -> Bool {
```
saeedbashir, line 93: You have renamed the `getComments` to `getResponsesData` but there are other elements which are according to the old method name like `self.comments`. Do we need to update them as well to it's ok to just update the method name?

saeedbashir:
@rnr The code is not compiling at my end and I am getting the following errors. I've tried by cleaning and even by deleting the derived data.
image

saeedbashir (approved):

volodymyr-chekyrta (approved):

Issue #391, opened 2024-04-08
[iOS] Completion doesn't work for videos in PiP mode.


sergeymomot:
Steps to reproduce:

1. open any content with video
2. complete all other cards (if they are present) except the card with video
3. go to the card with the video and start playing
4. switch to PiP mode
5. **exit the content (for example, to Profile or other content)**
6. wait for the video to finish playing
7. go to coursework and check completion status of the content
8. Result: content will not be completed, although all the cards have been completed and the video has been watched to the end

at the same time, if you switch to PiP mode and do not leave the content, then after playing the video to the end, the content will receive status "Completed".

PR #414, opened 2024-04-22
[iOS] Completion doesn't work for videos in PiP mode.

forgotvas:
Fix bug for:
[iOS] Completion doesn't work for videos in PiP mode. #391

A little refactor how players is build in.


shafqat-muneer:
@forgotvas I just noticed that even though the video is still playing in PiP mode, when I return from other screens, it appears as if the module has been completed.

https://github.com/openedx/openedx-app-ios/assets/11990137/9d5c517b-8fd5-43b1-9d60-77a33eb3633a

shafqat-muneer:
> @forgotvas I just noticed that even though the video is still playing in PiP mode, when I return from other screens, it appears as if the module has been completed.
> 
>  PiP.Mode.Video.mov

I came across this code snippet, and it appears to be functioning as intended, so there's no need to make any changes to it. 🎉 
```
if strongSelf.playerTracker.progress > 0.8 && !strongSelf.isViewedOnce {
                    strongSelf.isViewedOnce = true
                    Task {
                        await strongSelf.sendCompletion()
                    }
                }
```

shafqat-muneer:
@forgotvas I'm curious to understand what happens if a user downloads a video and watches it offline. Does the system recognize it as complete once the user reconnects online?

Course/Course/Presentation/Video/SubtittlesView.swift
shafqat-muneer, line 1: Typo: `SubtittlesView` to `SubtitlesView`

shafqat-muneer (changes requested):

forgotvas:
> @forgotvas I'm curious to understand what happens if a user downloads a video and watches it offline. Does the system recognize it as complete once the user reconnects online?

Hi @shafqat-muneer, you are right, it will try to call the completion request, but as we are offline it will not be send. Also in current implementation i don't see that we are syncing it after return to online mode. But that PR is fixing problem when we call completion for videos what you are playing in PIP mode, to solve offline sync we need to create another task & PR. @sergeymomot could create ticket for that.

 I have fixed code that you requested to change. Thank you for review.

shafqat-muneer:
> > @forgotvas I'm curious to understand what happens if a user downloads a video and watches it offline. Does the system recognize it as complete once the user reconnects online?
> 
> Hi @shafqat-muneer, you are right, it will try to call the completion request, but as we are offline it will not be send. Also in current implementation i don't see that we are syncing it after return to online mode. But that PR is fixing problem when we call completion for videos what you are playing in PIP mode, to solve offline sync we need to create another task & PR. @sergeymomot could create ticket for that.
> 
> I have fixed code that you requested to change. Thank you for review.

@forgotvas It makes sense to me. I've looked over the requested changes, and we are good to merge. 🎉

shafqat-muneer (approved):

shafqat-muneer (approved):

PR #450, opened 2024-06-07
fix: APIs path and archive fix

saeedbashir:
This PR address small issues with API paths and errors while creating archive.

1. Path issues:
Some API paths are starting without the `/`. Ideally it should work and it works sometimes because of `appendingPathComponent`.  But most of the time, it fails and path becomes `HostURLAPIPath` instead of `HostURL/APIPath`. Added the `/` to the API paths where it was missing.

2. Archive issue:
Some mock code was without the `#if DEBUG` and it was throughing an error while creating an archive. 

3. Theme
Some changes to fix the edX theme. These changes will not impact the openedX theme.

### DIFF
--- a/Core/Core/Network/AuthEndpoint.swift
+++ b/Core/Core/Network/AuthEndpoint.swift
@@ -29,9 +29,9 @@ enum AuthEndpoint: EndPointType {
         case .getAuthCookies:
             return "/oauth2/login/"
         case .getRegisterFields:
-            return "user_api/v1/account/registration/"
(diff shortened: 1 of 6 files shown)


volodymyr-chekyrta (approved):
LGTM 👍

volodymyr-chekyrta:
Please fix Lint Commit Message

rnr (approved):

forgotvas (approved):

PR #452, opened 2024-06-10
feat: Delete old downloaded videos data on device to optimize storage

shafqat-muneer:
[LEARNER-9966](https://2u-internal.atlassian.net/browse/LEARNER-9966): Delete old downloaded videos data on device to optimize storage

We decided not to migrate downloaded content but have to look into deleting the old downloaded videos so the app does not create useless dummy data in the learners' devices.

**Technical Details:**
The old application saves users' downloaded videos in the `Application Support` directory within folders named users' MD5 hashes. When multiple users log into the same device, separate MD5 hashed folders are created for each user, each containing that user's downloaded data.

After upgrading the application, we can use the MD5 hash algorithm to delete only the data of the currently logged-in user, while the data for other users will remain intact. Because MD5 hash folder name will be generated for current logged in username. To prevent this scenario, we need to create a regex that identifies and deletes all MD5 hashed folders upon launching the upgraded application.

We can consider removing this code once we are confident that most or all users have transitioned to the new application.

| Before Update Size | After Update Size |
|----------|----------|
| Before Update Size | After Update Size |

Before Update | After Update |
--- | --- | 
<video src="https://github.com/openedx/openedx-app-ios/assets/11990137/de782e20-a14b-4645-a363-24c6beae4cf6" /> | <video src="https://github.com/openedx/openedx-app-ios/assets/11990137/b5df90ca-e78a-4c17-8ca9-7a5bc9847c4d" /> |

### DIFF
--- a/Core/Core/Network/DownloadManager.swift
+++ b/Core/Core/Network/DownloadManager.swift
@@ -124,6 +124,8 @@ public protocol DownloadManagerProtocol {
     func resumeDownloading() throws
     func fileUrl(for blockId: String) -> URL?
     func isLargeVideosSize(blocks: [CourseBlock]) -> Bool
+    
+    func removeAppSupportDirectoryDeprecatedContent()
 }
 
 public enum DownloadManagerEvent {
@@ -470,6 +472,60 @@ public class DownloadManager: DownloadManagerProtocol {
             debugLog("SaveFile Error", error.localizedDescription)
         }
     }
+    
+    public func removeAppSupportDirectoryDeprecatedContent() {
+        deleteMD5HashedFolders()
+    }
+    
+    private func getApplicationSupportDirectory() -> URL? {
+        let fileManager = FileManager.default
+        do {
+            let appSupportDirectory = try fileManager.url(
+                for: .applicationSupportDirectory,
+                in: .userDomainMask,
+                appropriateFor: nil,
+                create: true
+            )
+            return appSupportDirectory
+        } catch {
+            print("Error getting Application Support Directory: \(error)")
+            return nil
+        }
+    }
+    
+    private func isMD5Hash(_ folderName: String) -> Bool {
+        let md5Regex = "^[a-fA-F0-9]{32}$"
+        let predicate = NSPredicate(format: "SELF MATCHES %@", md5Regex)
+        return predicate.evaluate(with: folderName)
+    }
+    
+    private func deleteMD5HashedFolders() {
+        guard let appSupportDirectory = getApplicationSupportDirectory() else {
+            return
+        }
+        
+        let fileManager = FileManager.default
+        do {
+            let folderContents = try fileManager.contentsOfDirectory(
+                at: appSupportDirectory,
+                includingPropertiesForKeys: nil,
+                options: []
+            )
+            for folderURL in folderContents {
+                let folderName = folderURL.lastPathComponent
+                if isMD5Hash(folderName) {
+                    do {
+                        try fileManager.removeItem(at: folderURL)
+                        print("Deleted folder: \(folderName)")
+                    } catch {
+                        print("Error deleting folder \(folderName): \(error)")
+                    }
+                }
+            }
+        } catch {
+            print("Error reading contents of Application Support directory: \(error)")
+        }
+    }
 }
 
 @available(iOSApplicationExtension, unavailable)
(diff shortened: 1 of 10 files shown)


Core/Core/Network/DownloadManager.swift
```diff
@@ -124,6 +124,8 @@ public protocol DownloadManagerProtocol {
     func resumeDownloading() throws
     func fileUrl(for blockId: String) -> URL?
     func isLargeVideosSize(blocks: [CourseBlock]) -> Bool
+    
+    func removeAppSupportDirectoryDeprecatedContent()
```
saeedbashir, line 128: How about renaming it to `removeAppSupportDirectoryUnusedContent` thoughts?

Core/Core/Network/DownloadManager.swift
```diff
@@ -470,6 +472,60 @@ public class DownloadManager: DownloadManagerProtocol {
             debugLog("SaveFile Error", error.localizedDescription)
         }
     }
+    
+    public func removeAppSupportDirectoryDeprecatedContent() {
+        deleteMD5HashedFolders()
```
saeedbashir, line 477: How about doing the work directly here that you are doing in `deleteMD5HashedFolders` function?

Core/Core/Network/DownloadManager.swift
```diff
@@ -470,6 +472,60 @@ public class DownloadManager: DownloadManagerProtocol {
...
+        do {
+            let appSupportDirectory = try fileManager.url(
+                for: .applicationSupportDirectory,
+                in: .userDomainMask,
+                appropriateFor: nil,
+                create: true
+            )
+            return appSupportDirectory
+        } catch {
+            print("Error getting Application Support Directory: \(error)")
```
saeedbashir, line 491: Please use debugPrint here and in other places where applicable in code that you wrote.

OpenEdX/RouteController.swift
```diff
@@ -99,4 +100,22 @@ class RouteController: UIViewController {
...
+    private func resetAppSupportDirectoryUserData() {
+        if var upgradationValue = Container.shared.resolve(CoreStorage.self) {
+            if upgradationValue.resetAppSupportDirectoryUserData == false {
+                Task {
+                    Container.shared.resolve(DownloadManagerProtocol.self)?.removeAppSupportDirectoryDeprecatedContent()
+                    upgradationValue.resetAppSupportDirectoryUserData = true
+                }
+            }
+        }
+    }
```
saeedbashir, line 120: How about updating this method to guard like below thoughts?
```
private func resetAppSupportDirectoryUserData() {
        guard var upgradationValue = Container.shared.resolve(CoreStorage.self),
              let downloadManager = Container.shared.resolve(DownloadManagerProtocol.self),
              upgradationValue.resetAppSupportDirectoryUserData == false
        else {return}
        
        Task {
            downloadManager.removeAppSupportDirectoryDeprecatedContent()
            upgradationValue.resetAppSupportDirectoryUserData = true
        }
    }

File: Core/Core/Network/DownloadManager.swift [hunk 5]
```diff
@@ -470,6 +472,60 @@ public class DownloadManager: DownloadManagerProtocol {
...
+            return nil
+        }
+    }
+    
+    private func isMD5Hash(_ folderName: String) -> Bool {
+        let md5Regex = "^[a-fA-F0-9]{32}$"
```
rnr, line 497: Can md5 contain uppercase letters?

rnr (approved):
looks good to me (just check Saeed's comments)

Core/Core/Network/DownloadManager.swift
shafqat-muneer: For our specific scenario, the suggested name `removeAppSupportDirectoryUnusedContent` seems fitting. However, the name `deleteMD5HashedFolders` explicitly describes the deletion of MD5 hashed folders (not any other content). If we integrate the implementation of `deleteMD5HashedFolders()` directly into `removeAppSupportDirectoryUnusedContent()`, it might affect the clarity of the abstraction level. Do you have any suggestions or thoughts on finding a more suitable name?

Core/Core/Network/DownloadManager.swift
shafqat-muneer: True, an MD5 hash will not contain uppercase letters. However, to be safe, if there is any customization that introduces uppercase letters, it will also delete that folder.

saeedbashir (approved):
LGTM 👍

shafqat-muneer:
@volodymyr-chekyrta Awaiting your review. Thanks

Issue #436, opened 2024-05-13
[iOS] Unexpected exit from the course


sergeymomot:
Steps to repeat:
1. Open a course with video
2. Navigate to Video tab
3. Open "Video download quality" page
4. Tap Back arrow to close "Video download quality" page
5. Result: an unexpected exit from the course occurs

Expected result: after tap on Back arrow only "Video download quality" page should close

https://github.com/openedx/openedx-app-ios/assets/2964930/8668cf91-7298-484b-a2ba-88a0ca9f9cbd

PR #457, opened 2024-06-11
Fix video quality settings in modal representation

rnr:
This PR is the fix for https://github.com/openedx/openedx-app-ios/issues/436
In a modal view, the modal representation of that view should be used.
Added `isModal` parameter with a default value of `false`, which determines which presentation will be used.
The full screen view  in the Settings does not change.

Now | After fix
--- | ---
 |


volodymyr-chekyrta (approved):

shafqat-muneer (approved):
LGTM 👍

forgotvas (approved):

PR #518, opened 2024-09-06
fix: fix coredata crash on primary course

rnr:
This PR fixes the crash with coredata. Although, I'm not able to reproduce the crash  but from dry run it it can occur because of deleting the data async and right before saving new data and expecting the data will be deleted. 

Crash Link: https://console.firebase.google.com/u/1/project/openedx-mobile/crashlytics/app/ios:org.edx.mobile/issues/cb09ae273338d6fa32acb9f64effefbf?time=last-seven-days&types=crash&versions=6.0.0%20(2)&sessionEventKey=f4c972ede3fd4c18901af18a71e6b702_1979319760445579771

Screenshot 2024-08-09 at 8 03 46 AM

### DIFF
--- a/OpenEdX/Data/DashboardPersistence.swift
+++ b/OpenEdX/Data/DashboardPersistence.swift
@@ -231,7 +231,7 @@ public class DashboardPersistence: DashboardPersistenceProtocol {
     // swiftlint:enable function_body_length
     
     func clearOldEnrollmentsData() {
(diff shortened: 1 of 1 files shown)


volodymyr-chekyrta (approved):
LGTM 👍 
The best way to handle this would probably be to convert it to an async await function, but we can do that later for the whole class.

PR #519, opened 2024-09-06
fix: core data crash with inverse attribute

rnr:
This PR fixes crash:

IMG_7707


volodymyr-chekyrta:
@IvanStepanok will review this, because deletionRule=“Cascade” was done intentionally to fix a different crash case.

rnr:
Thank you @IvanStepanok !
I tried this and looks good. Lets wait opinion from @forgotvas (he had fixed this crash)

forgotvas:
hi @IvanStepanok, yes main point is to add relationship between `CDPrimaryCourse` and `CDAssignment`, `Nulify` rule was committed by my fault, so our fix should work fine with `cascade` rule too. Like you suggested in your snapshot, but you did use new fields for relationship.

Can you provide example what crash fixed by `Cascade` rule?

forgotvas:
> Hi @forgotvas
> 
> We added a cascading deletion rule to fix an issue where reloading CDPrimaryCourse did not overwrite nested objects properly. By using "Cascade," related entities (like CDAssignment) are now deleted alongside their parent, ensuring that when reloading, new data can be saved without conflicts from residual old data.

do you have steps to reproduce that crash, want to try to repeat it in my branch with my fix.

IvanStepanok:
> do you have steps to reproduce that crash, want to try to repeat it in my branch with my fix.

As I remember, you need a really large course to reproduce this error. On the first attempt, caching works fine, but on the second time, there's a cascade of core data errors. Nothing special, the app still works, but the cache is not updated.

rnr:
So @IvanStepanok @forgotvas 
As I see there is no difference between your fixes - just Ivan creates separate Relationships to relates entities but Vadim add `inverseEntity` field. Is this right? Then lets select one of these fixes and move forward. WDYT?

volodymyr-chekyrta:
Please resolve git conflicts

rnr:
Conflict is resolved
Also changed deletionRules to 'Cascade' as @IvanStepanok suggested

forgotvas (approved):

volodymyr-chekyrta (approved):
LGTM

PR #575, opened 2025-03-03
v1.0.3 plugins support

rnr:
This PR contains the following changes:
- did set version 1.0.3 for `OEXFoundation` and `OEXFirebaseAnalytics`.
- removed `BrazeProvider` and `BrazeListener`
- cleaned up `PushNotificationsManager` 
- added `pushNotificationsProviders` and `pushNotificationsListeners` to `PluginManager` - now we can add here any services implemented via plugin. Now we can move FCM to a separate plugin if we want.


volodymyr-chekyrta (approved):
Looks good to me 👍

PR #664, opened 2026-07-08
feat: LMS Directory — let a learner pick which Open edX platform to sign in to

IvanStepanok:
## What this adds

A learner opens the app and picks which Open edX platform to sign in to, instead
of the app being built for exactly one site. The app then re-themes to the chosen
platform — its logo, accent colour and sign-in artwork — and authenticates
against it.

Behind `LMS_DIRECTORY.ENABLED`, **off by default**. With the flag off nothing
here runs and the app behaves exactly as it does today; every default config in
this PR ships with the feature disabled and no URL set.

## Where the platform list comes from

Three ways, and the config alone decides which:

```yaml
LMS_DIRECTORY:
  ENABLED: true
  DIRECTORY_URL: "https://example.com/lms_directory.json"   # a document, on the web
  DIRECTORY_FILE: ""                                        # or: a document, in the app
  DIRECTORY_MODE: ""
```

A `DIRECTORY_URL` ending in `.json` is read as a **document**: one file, fetched
once, that already contains every platform and its branding. Any other address is
treated as a **live registry** answering `/api/v1/directory`. A file bundled with
the app wins over both — a build that ships its own copy has deliberately opted
out of the network.

The document is the part worth reviewing carefully, because it is what makes this
usable without anyone running a service. It also removed a real hazard: both apps
used to fall back to open search when a registry was unreachable, so a branded
build starting offline showed a search box over every public platform. A document
has no server to ask, so there is nothing to fall back to.

Format, both delivery modes and a worked example: **[`Documentation/LMS_DIRECTORY.md`](Documentation/LMS_DIRECTORY.md)**.

## Reporting a platform is not in this PR's scope

"Report this LMS" appears only when the directory is a live registry that says it
is an open catalog — the one situation where a stranger can list something nobody
has vouched for. A build reading a document never shows it.

That decision is enforced by an explicit three-state mode — `unknown | search |
curated` — rather than a boolean. A boolean has no room for "the server has not
said yet", so its absent value had to mean something, and it meant *open*: an
upgraded install and a build newly pointed at another registry would both offer
reporting for a catalog nobody had vouched for. The mode is stored against a key
naming the source that produced it, refreshed on every launch (a learner who has
picked a platform never sees the picker again, so a registry that changes its mode
would otherwise go unnoticed), and anything unknown stays hidden.

## Testing

`CoreTests` and `AuthorizationTests`: **103 tests, 0 failures** on an iPhone 17
Pro simulator. They cover reading a document from a URL and from the bundle,
image sources resolving to either an address or a bundled name, and every
transition of the directory mode — including an upgrade from a build that stored
only a boolean, and the same registry changing its mode at the same address.

`LMSDirectoryUITests` drives the running app signed in and asserts on whether the
reporting entry point is there. It skips itself unless the environment configures
it (`TEST_RUNNER_LMS_DIRECTORY_UITEST=1`, an expectation, and credentials), since
it needs something to sign in to; the header explains what to point it at.

Verified on a simulator, signed in, both ways round: a build reading a document
offers no reporting, and the same install rebuilt against a registry — data kept,
platform picker skipped — offers it again.

## Notes for reviewers

- `Core` gained `LMSDirectoryState`, which owns the whole rule about what the
  directory is and how long that knowledge is good for. `ProfileView` observes it
  so the entry point appears when the answer arrives rather than on a later visit.
- `Theme.Images.headerBackground` holds a decoded `UIImage` rather than a URL.
  The sign-in header used `AsyncImage`, which has no cache, so it refetched and
  flashed a placeholder on every appearance. Kingfisher does the fetching;
  `Theme` does not depend on it, because `Core` already depends on `Theme`.
- Image fields carry either an `http(s)` address or the name of a file in the
  bundle. One field, one rule — the alternative was a parallel set of `*_asset`
  fields and a precedence rule to go with them.


openedx-webhooks:
Thanks for the pull request, @IvanStepanok!


This repository is currently maintained by `@openedx/openedx-mobile-maintainers`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.
</details>

<details><summary>:radio_button: Provide context</summary>

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads
</details>





<details><summary>:radio_button: Get a green build</summary>

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.
</details><details>


---

<details><summary>Where can I find more information?</summary>

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)
</details>

<details><summary>When can I expect my changes to be merged?</summary>

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*
</details>

PR #670, opened 2026-08-20
feat: multi-tenant — let a learner choose which Open edX platform to sign in to

IvanStepanok:
## What this adds

A build of this app normally talks to exactly one Open edX site, named in
`config.yaml`. With `LMS_DIRECTORY` enabled it instead shows a list of platforms,
lets the learner pick one, re-themes to it — logo, accent colour, sign-in artwork
— and signs in against that one. Logging out returns to the list, so a device can
move between platforms without a new build.

**Off by default.** With `ENABLED: false` none of this runs and the app behaves
exactly as it does today. Every shipped config in this PR has it disabled with no
source set.

## Where the list comes from

One JSON document, given either way — never both:

```yaml
LMS_DIRECTORY:
  ENABLED: true
  DIRECTORY_URL: "https://example.com/lms_directory.json"   # fetched once
  DIRECTORY_FILE: ""                                        # or: "lms_directory.json" in the app target
```

A bundled file wins over an address: a build that ships its own copy has
deliberately opted out of the network, and quietly preferring a remote list would
undo that. Image fields inside the document are either web addresses or names of
files shipped with the app, so a build can be fully offline — list and artwork
both — and still be branded.

Because the whole list arrives at once, every platform's sign-in background is
known before anything is tapped. That is what makes it possible to warm the
artwork while the learner is still choosing, instead of watching the branded
screen assemble itself afterwards.

Format, both delivery modes and a worked example: **[`Documentation/LMS_DIRECTORY.md`](Documentation/LMS_DIRECTORY.md)**.

## Scope

Multi-tenancy only: read a list, choose, sign in, log out, choose again. No
search, no QR sign-in, no reporting a platform, and nothing that talks to a
catalog service — the app knows about a document and nothing else. Those belong
to a universal-app phase and will come as their own PRs.

## Testing

`CoreTests` and `AuthorizationTests`: **75 tests, 0 failures** on an iPhone 17 Pro
simulator. They cover which source a config resolves to, reading a document from
a URL and from the bundle (including a decoding failure, an HTTP error and a lost
connection), image sources resolving to either an address or a bundled name, the
picker's states, and a selection reaching the coordinator that re-themes the app.

One of them pins the smallest document a person could reasonably write by hand,
because the same file has to work on Android too.

## Notes for reviewers

- `Theme.Images.headerBackground` holds a decoded `UIImage` rather than a URL.
  The sign-in header used `AsyncImage`, which has no cache, so it refetched and
  flashed a placeholder on every appearance. Kingfisher does the fetching;
  `Theme` does not depend on it, because `Core` already depends on `Theme`.
- `LMSSelectionCoordinator` is built inline on the main actor rather than
  registered with Swinject: the coordinator is `@MainActor`-isolated while
  Swinject's factory closure is not, and a registration drops the global actor.

### DIFF
--- a/Authorization/Authorization/Presentation/TenantPicker/LMSOverridesStore.swift
+++ b/Authorization/Authorization/Presentation/TenantPicker/LMSOverridesStore.swift
@@ -1,0 +1,79 @@
+import Core
+import Foundation
+
+struct LMSSelectionSnapshot: Codable, Sendable {
+    let detail: LMSDetailDTO
+    let appliedAt: Date
+}
+
+protocol LMSOverridesStoreProtocol: Sendable {
+    func save(detail: LMSDetail, payload: Data, storage: CoreStorage?) throws
+    func currentSelection() -> LMSDetail?
+    func clear(storage: CoreStorage?) throws
+}
+
+final class LMSOverridesStore: LMSOverridesStoreProtocol {
+    enum Keys {
+        static let selectionPayload = "lmsDirectory.selected_lms_payload"
+        static let feedbackEmail = "lmsDirectory.selected_feedback_email"
+        static let oauthClientId = "lmsDirectory.selected_oauth_client_id"
+        static let accentColor = "lmsDirectory.selected_accent_color"
+        static let accentColorDark = "lmsDirectory.selected_accent_color_dark"
+        static let unknownUnitsMode = "lmsDirectory.selected_unknown_units_mode"
+        static let loginBackgroundURL = "lmsDirectory.selected_login_background_url"
+        static let logoUploadURL = "lmsDirectory.selected_logo_upload_url"
+        static let courseUnitProgress = "lmsDirectory.selected_course_unit_progress"
+        static let courseDropdownNav = "lmsDirectory.selected_course_dropdown_nav"
+        static let preLoginExperience = "lmsDirectory.selected_pre_login_experience"
+        static let dashboardType = "lmsDirectory.selected_dashboard_type"
+    }
+
+    private nonisolated(unsafe) let userDefaults: UserDefaults
+
+    init(userDefaults: UserDefaults = .standard) {
+        self.userDefaults = userDefaults
+    }
+
+    func save(detail: LMSDetail, payload: Data, storage: CoreStorage?) throws {
+        if var storage = storage {
+            storage.selectedLMSBaseURL = detail.api.hostURL.absoluteString
+        }
+        userDefaults.set(payload, forKey: Keys.selectionPayload)
+        userDefaults.set(detail.api.feedbackEmail, forKey: Keys.feedbackEmail)
+        userDefaults.set(detail.api.oauthClientId, forKey: Keys.oauthClientId)
+        userDefaults.set(detail.accentColorHex, forKey: Keys.accentColor)
+        userDefaults.set(detail.theme?.accentColorDark, forKey: Keys.accentColorDark)
+        userDefaults.set(detail.featureFlags.unknownUnitsMode ?? "block", forKey: Keys.unknownUnitsMode)
+        userDefaults.set(detail.theme?.loginBackgroundURL?.absoluteString, forKey: Keys.loginBackgroundURL)
+        userDefaults.set(detail.effectiveLogoURL?.absoluteString, forKey: Keys.logoUploadURL)
+        userDefaults.set(detail.uiComponents?.courseUnitProgressEnabled ?? true, forKey: Keys.courseUnitProgress)
+        userDefaults.set(detail.uiComponents?.courseDropdownNavigationEnabled ?? true, forKey: Keys.courseDropdownNav)
+        userDefaults.set(detail.uiComponents?.preLoginExperienceEnabled ?? true, forKey: Keys.preLoginExperience)
+        userDefaults.set(detail.dashboard?.type ?? "gallery", forKey: Keys.dashboardType)
+    }
+
+    func currentSelection() -> LMSDetail? {
+        guard
+            let payload = userDefaults.data(forKey: Keys.selectionPayload),
+            let dto = try? JSONDecoder().decode(LMSDetailDTO.self, from: payload)
+        else {
+            return nil
+        }
+        return dto.domainModel
+    }
+
+    func clear(storage: CoreStorage?) throws {
+        if var storage = storage {
+            storage.selectedLMSBaseURL = nil
+        }
+        for key in [
+            Keys.selectionPayload, Keys.feedbackEmail, Keys.oauthClientId,
+            Keys.accentColor, Keys.accentColorDark, Keys.unknownUnitsMode,
+            Keys.loginBackgroundURL, Keys.logoUploadURL,
+            Keys.courseUnitProgress, Keys.courseDropdownNav,
+            Keys.preLoginExperience, Keys.dashboardType
+        ] {
+            userDefaults.removeObject(forKey: key)
+        }
+    }
+}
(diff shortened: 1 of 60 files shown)


openedx-webhooks:
Thanks for the pull request, @IvanStepanok!


This repository is currently maintained by `@openedx/openedx-mobile-maintainers`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.
</details>

<details><summary>:radio_button: Provide context</summary>

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads
</details>





<details><summary>:radio_button: Get a green build</summary>

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.
</details><details>


---

<details><summary>Where can I find more information?</summary>

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)
</details>

<details><summary>When can I expect my changes to be merged?</summary>

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*
</details>

IvanStepanok:
To test this MR you can create a JSON file by using my service: https://providers.openedx-lms.stepanok.com
Or use this JSON file: https://providers.openedx-lms.stepanok.com/p/northwind-education-group/directory.json (some of LMS in this JSON file are disabled)

edschema:
Looks great: one UX piece of feedback: there's no good way to get back to the LMS site selection page if the selected site has pre-login discovery. If a learner accidentally clicks the wrong LMS that has pre-login discovery enabled, the learner has to click "sign in" and then click "change" in the selected LMS block in order to get back to the site selector. 

I'm logging minor notes as well, I'll share once I've finished testing

RawanMatar89:
@IvanStepanok Sorry for the slow turnaround here — this has been sitting longer than it should have. I've started digging into it and will post a full review today or tomorrow at the latest. Thanks for your patience! :)

Authorization/Authorization/Presentation/TenantPicker/LMSSelectionCoordinator.swift
```diff
@@ -0,0 +1,75 @@
...
+            LMSThemeApplier.applyAccentColor(detail.accentColor, darkColor: detail.accentColorDark)
+            LMSThemeApplier.applyLoginBackground(LMSImageSource(url: detail.theme?.loginBackgroundURL))
+            reRegisterAPI(with: detail.api.hostURL)
+            await handlePostSelection(for: detail)
+        } catch {
+            assertionFailure("Failed to apply LMS selection: \(error)")
+        }
+    }
+
+    private func reRegisterAPI(with baseURL: URL) {
```
RawanMatar89, line 62: Since preLoginDiscovery can present the Discovery screen before sign-in, could you walk me through this sequence: 
pick platform A -> browse Discovery (which resolves and caches an API-backed repository pre-login) -> tap Sign -in -> tap "Change" -> pick platform B. 
Does Discovery now hit B, or does the repository resolved during A's session keep the old API instance (and host) since container-scope re-registration only affects objects resolved after the swap.

If that's a real gap, one pattern worth considering: instead of swapping the DI registration, resolve the base host per-request inside the RequestInterceptor/adapter, reading the current selection at request time rather than at object-construction time. That way nothing can hold a stale host reference, regardless of when it was built. Not asking you to redo the DI approach wholesale if I'm wrong about the gap above — just want to make sure this is tested either way, since it's the kind of bug that only shows up after a "Change platform" tap mid-session, not on first launch.

Authorization/Authorization/Presentation/TenantPicker/StaticLMSDirectoryService.swift
```diff
@@ -0,0 +1,197 @@
...
+    /// Every image the app will need, so a caller can warm them before they are shown.
+    func imageSources() async throws -> [LMSImageSource] {
+        try await allPlatforms().flatMap(\.imageSources)
+    }
+
+    // MARK: - Private
+
+    private func allPlatforms() async throws -> [LMSDetail] {
+        try await document().include.enumerated().map { index, entry in
+            entry.domainModel(id: String(index))
```
RawanMatar89, line 103: this assigns id: String(index) from array position rather than a stable identifier from the document — flagged in your own last commit. If the directory's order isn't guaranteed stable across fetches, a stale in-flight selection could resolve to the wrong platform. Worth confirming the format guarantees stable ordering, or exposing a real id field.

Authorization/Authorization/Presentation/TenantPicker/LMSOverridesStore.swift
```diff
@@ -0,0 +1,88 @@
...
+    let appliedAt: Date
+}
+
+protocol LMSOverridesStoreProtocol: Sendable {
+    func save(detail: LMSDetail, payload: Data, storage: CoreStorage?) throws
+    func currentSelection() -> LMSDetail?
+    func clear(storage: CoreStorage?) throws
+}
+
+final class LMSOverridesStore: LMSOverridesStoreProtocol {
```
RawanMatar89, line 22: The four Config.swift overrides (baseURL, baseSSOURL, oAuthClientId, feedbackEmail) and LMSOverridesStore's twelve keys hit UserDefaults.standard directly with hand-written string keys, and the same key strings are re-typed independently in Config.swift, LMSOverridesStore.Keys, and ConfigLMSDirectoryTests.swift. A typo or rename in one place would silently break the others with no compiler error. Since this PR already adds CoreStorage.selectedLMSBaseURL as a proper DI-testable property, would it make sense to route all of these overrides through one injectable store, or at minimum share the key constants, rather than three independent copies? (e.g. "lmsDirectory.selected_oauth_client_id")

RawanMatar89 (commented):
Thank you @IvanStepanok for this — the scope of what's covered (offline bundled catalog, dynamic theming, the test coverage) is genuinely impressive 👏🏻 A few things I'd like to talk through that may affect the design:

